PR #21, opened 2019-11-26
Implement file storage with an item pipeline

oltarasenko:
This commit introduces a change in a general Crawly Data Storage Worker.
Now the worker is not responsible for storing the item itself. And
this work is expected to be done by a pipeline.

Also file_storage_pipeline is added.


oltarasenko:
Todo:

- [ ] Fix coverage tests (FilesPipeline is not covered yet)
- [ ] Update docsify documentation (deprecated settings)
- [ ] Write a comment why we're not doing the File.close in the pipeline, so it's clear to readers

Ziinc:
I've renamed the pipeline to `WriteToFile`, updated relevant docs sections. I've also written a simplified unit test with file assertions.

oltarasenko:
👍 Still need to do something with coveralls. Maybe we should decrease the required coverage a bit. I think we could try to aim 85% for example?

Ziinc:
i agree, coverage != good code

oltarasenko:
@Ziinc now the build is passing. So if it's ok with you lets merge it?

Ziinc:
The use of global configs for pipelines makes me a little uncomfortable, but i think that's for another issue/PR.

I fixed a typo; looks good to merge

PR #23, opened 2019-11-29
Removed Floki as a dependency. It was outdated and unused.

jallum:


oltarasenko:
Good catch as far as I see it. I will test it a bit later today, to double everything and will merge it!

Ziinc:
Need to update tutorial to add floki as a dep.

Issue #22, opened 2019-11-27
Move away from using global app env for pipeline module config


Ziinc:
Pipeline config should be localized to the specific pipeline. 
Benefits:
1. Reduces clashes (when more middlewares/pipelines get built)
2. makes parameter declaration clearer
3. makes pipelines more "functional" and reusable
4. use of `Application.get_env` usage within pipeline modules makes things less clear when declaring pipelines.

Related to #20 , would pave the way for adding logic into a pipeline (instead of having a fat pipeline module)

Proposed api:
- Tuple definitions would only be required for modules that require configuration
    - at the pipeline level, they could throw an error when checking for parameters, or use default values.
- a pipeline module can also be passed directly when no parameters are needed

For example:
```elixir
pipelines: [
  ....
  MyCustom.Pipeline.CleanItem,
  {Crawly.Pipelines.Validate, item: [:title, :url] },
  {Crawly.Pipelines.DuplicatesFilter, item_id: :title },
  Crawly.Pipelines.JSONEncoder
]
```
Besides adjusting existing built-in pipelines, this proposed change would also require the adjustment of `Crawly.Utils.pipe` to check for tuple definitions.

oltarasenko:
I kind of like this idea!

Ziinc:
I'll open a PR for this over the weekend

Issue #20, opened 2019-11-25
Scraping different items from the same spider, each with different pipeline requirements


Ziinc:
Problem:
Crawly only allows single item type scraping. However, what if i am crawling two different sites, with vastly different items? 

For example, web page A (e.g. a blog) will have:
- comments
- article content
- related links
- title

while web page B (e.g. a weather site) will have:
- temperature
- country

In the current setup, the only way to work around this is to lump all these logically different items into one large item, such that the end item declaration in config will be:
```
item: [:title, :comments, :article_content, :related_links, :temperature, :country]
```

the issues are that:
1. they share the same pipeline. When scraping the weather data, the blog-related fields will be blank, and vice versa for when scraping the blog. This will affect pipeline validations, since pipeline is shared.
2. Output of the two items will be the same.
3. Duplication checks (such as `:item_id`) is not item specific. Since the item-type from the weather site has no title, I can't specify an item-type-specific field.

---

I have some idea of how this could be implemented, taking inspiration from scrapy.
We could define item structs, and sort the items to their appropriate pipelines according to struct.

Using the tutorial as an example:

using this ideal scenario config:
```
config :crawly,
  closespider_timeout: 10,
  concurrent_requests_per_domain: 8,
  follow_redirects: true,
  closespider_itemcount: 1000,
  middlewares: [
    Crawly.Middlewares.DomainFilter,
    Crawly.Middlewares.UniqueRequest,
    Crawly.Middlewares.UserAgent
  ],
  pipelines: [
    {MyItemStruct, [
        Crawly.Pipelines.Validate,
        {Crawly.Pipelines.DuplicatesFilter, item_id: :title }, # similar to how supervisor trees are declared
        Crawly.Pipelines.CSVEncoder
    ]},
    {MyOtherItemStruct, [
        Crawly.Pipelines.Validate,
        Crawly.Pipelines.CleanMyData,
        {Crawly.Pipelines.DuplicatesFilter, item_id: :name }, # similar to how supervisor trees are declared
        Crawly.Pipelines.CSVEncoder
    ]},
  ]

```

with the spider implemented as so:
```elixir
 @impl Crawly.Spider
  def parse_item(response) do
    hrefs = response.body |> Floki.find("a.more") |> Floki.attribute("href")

    requests =
      Utils.build_absolute_urls(hrefs, base_url())
      |> Utils.requests_from_urls()

    title = response.body |> Floki.find("article.blog_post h1") |> Floki.text()
[...]

PR #28, opened 2019-11-29
A sketch of how per-spider pipelines could work

jallum:
This would allow a pipeline config that looks like this:
```
  pipelines: %{
    Some.Spider.ForASpecificSite => [
      Crawly.Pipelines.Validate,
      Crawly.Pipelines.JSONEncoder,
      Crawly.Pipelines.WriteToFile
    ],
    default: [
      Crawly.Pipelines.Validate,
      Crawly.Pipelines.DuplicatesFilter,
      Crawly.Pipelines.JSONEncoder,
      Crawly.Pipelines.WriteToFile
    ]
  }
```

...while allowing a fallback to the existing one-size-fits all pipeline list. The "default:" entry is optional, and would be used if no spider-specific entry is found in the map (otherwise, [] as now.)

### DIFF
--- a/lib/crawly/data_storage/data_storage_worker.ex
+++ b/lib/crawly/data_storage/data_storage_worker.ex
@@ -14,7 +14,7 @@ defmodule Crawly.DataStorage.Worker do
 
   use GenServer
 
-  defstruct stored_items: 0, spider_name: nil
+  defstruct stored_items: 0, spider_name: nil, pipeline: nil
 
   def start_link(spider_name: spider_name) do
     GenServer.start_link(__MODULE__, spider_name: spider_name)
@@ -29,14 +29,17 @@ defmodule Crawly.DataStorage.Worker do
   end
 
   def init(spider_name: spider_name) do
-    {:ok, %Worker{spider_name: spider_name}}
+
+    pipeline = pipeline_from_config(spider_name)
+
+    {:ok, %Worker{spider_name: spider_name, pipeline: pipeline}}
   end
 
   def handle_cast({:store, item}, state) do
-    pipelines = Application.get_env(:crawly, :pipelines, [])
-
+    %{ pipeline: pipeline } = state
+    
     state =
-      case Crawly.Utils.pipe(pipelines, item, state) do
+      case Crawly.Utils.pipe(pipeline, item, state) do
         {false, new_state} ->
           new_state
 
@@ -50,4 +53,16 @@ defmodule Crawly.DataStorage.Worker do
   def handle_call(:stats, _from, state) do
     {:reply, {:stored_items, state.stored_items}, state}
   end
+
+  defp pipeline_from_config(spider_name) do
+    pipeline_from_config_entry(Application.get_env(:crawly, :pipelines, []), spider_name) || []
+  end
+
+  defp pipeline_from_config_entry(%{} = mappings, spider_name) do
+    case mappings |> Map.get(spider_name) || mappings |> Map.get(:default) do
+      nil -> []
+      list when is_list(list) -> list
+    end
+  end
+  defp pipeline_from_config_entry(list, _spider_name) when is_list(list), do: list
 end


oltarasenko:
**[I am just thinking out loudly]**

I just wonder what if we define these on spider level instead? E.g. from one side, we will have more simple config, and spider would kind of guess what should be done....

however, from the other side. Do we want to do write to file inside every spider? Should spiders know where to send the data, etc?

Ziinc:
A spider could be scraping multiple data types from a page.

I suppose that there could be source-specific data cleaning steps, but no reason why it can't be implemented as a conditional step in a pipeline.

As mentioned in #20, a possible method is struct-based filtering or conditional pipeline logic (or a combinationoof both).

@jallum 's example could also be implemented as a Case pipeline, matching elixir's `case` behaviour. Less new things to learn too.

Ziinc:
From my point of view, a set of pipelines should suit the scraped item, since the cleaning methods would be specific to that data structure.

The spider doesn't have to know what happens afterwards, it's only concerned with the extraction. The post-processing is an entirely separate phase.

An analogy is photography. The image capture process may be different for each photo, but the the post-processing would depend on what type of image you intend to produce.

Ziinc:
> As mentioned in #20, a possible method is struct-based filtering or conditional pipeline logic (or a combinationoof both).
> 
> @jallum 's example could also be implemented as a Case pipeline, matching elixir's `case` behaviour. 

Something like this (with #22 resolved) :
```elixir
pipelines: [
    ...,
    {Crawly.Pipelines.Logic, func: case do
        {_item, %{spider_name: s} = state} when s == MySpiderName -> [ MySpiderSpecificPipeline ]
        _ -> [ MyDefaultPipeline ]
    end},
    Crawly.Pipelines.WriteToFile
]
```
where the `Crawly.Pipelines.Logic` receives a function as a parameter that dynamically decides the subsequent pipelines based on the item-state tuple it receives.

This makes it more flexible, as by using the using the item-state tuple:
1. we can check on where the item came from (using state.spider_name) 
2. we can perform checks on the item (thereby resolving #20)
3. non-breaking change (would just need #22 implemented, which is backwards compatible)
4. simple to ship (would simply be a recursive call to `Crawly.Utils.pipe`, would only need 1 new pipeline module to be maintained)
5. generalized enough to accept any function that returns a pipeline declaration list, making it more flexible for other use/edge cases.
6. can possibly be refined further for use with middlewares

oltarasenko:
Now again, why would we need a separate pipelines set for a spider after all?

E.g. a pipeline is a generic way of working with multiple spiders. E.g. if I want to generate uuids based on some fields, or drop items without prices everywhere, I would use pipeline. So the spider code does not know about it.

I think the main concern is related to a question:  Do we need per spider configuration of the pipelines?

jallum:
It's a fair question. In my use case, I have a good number of spiders that produce a single type of item. I'm anticipating that I'll soon have need for another group of spiders that produce something different. I'll need to process those items differently, and route them to a different place. So, I guess I could restate the question as, "Would it be useful to be able to process different _types of items_ with different pipelines?"

Asked that way, I think that tweaking my suggestion above to key off the type of struct produced (instead of the spider that produced it) would maybe make more sense.

Ziinc:
> "Would it be useful to be able to process different types of items with different pipelines?"
This is also the issue i am facing (as already mentioned in #20. Right now I'm experimenting with the key denoting the data type, and the nested value as the scraped item and filtering
```elixir
item = %{
    weather: %{ weather_data: "bla" }
    article: %{ article_data: "bla" }
}

# in custom pipeline
case item do
    %{weather: weather} -> ... # clean weather data
    %{article: article -> ... # clean article data
    %{comments: comments-> ... # clean comments data
end
```
This approach results in:
- fat pipeline module
- non-reusable logic
- single pipeline module

and my config becomes just a single declaration, which seems counter-intuitive to what a pipeline is for:
```elixir
pipelines: [MyFatPipeline]
```

Ziinc:
@jallum Refer to this comment on item-based pipeline definitions

oltarasenko:
@Ziinc Can I also know more about your case? E.g. I want to keep the code simple... (ideally). Maybe I will be able to advise something if I will play with your problem a bit. (Maybe lets proceed in a ticket, as the PR should stay clean of non code related conversation)

Ziinc:
Closing for now. api discussion to continue in #20

Issue #24, opened 2019-11-29
Improve tests for CSV encoder pipeline


oltarasenko:
Some tests are quite ugly at the moment :(. Need to check the race condition here:

```
  1) test CSV encoder test Items are stored in CSV after csv pipeline (DataStorageWorkerTest)
     test/data_storage_worker_test.exs:149
     ** (MatchError) no match of right hand side value: {:error, :already_started}
     stacktrace:
       test/data_storage_worker_test.exs:6: DataStorageWorkerTest.__ex_unit_setup_0/1
       test/data_storage_worker_test.exs:1: DataStorageWorkerTest.__ex_unit__/2
```

As it fails the CI pipeline for master.

Ziinc:
I'll handle standardizing the pipeline tests

PR #29, opened 2019-11-30
standardized existing pipeline testing

Ziinc:
For issue #24 .


test/pipelines/duplicates_filter_test.exs
```diff
@@ -0,0 +1,40 @@
+defmodule Pipelines.DuplicatesFilterTest do
+  use ExUnit.Case, async: false
+
+  @valid %{data: [%{some: "nested_data"}], id: "my_id"}
+
+  test "DuplicatesFilter prevents drops duplicate items with the same item_id value" do
+    Application.put_env(:crawly, :item_id, :id)
```
oltarasenko: CI returns an error here:
```
  1) test DuplicatesFilter is inactive when item_id is not set (Pipelines.DuplicatesFilterTest)
     test/pipelines/duplicates_filter_test.exs:23
     Assertion with == failed
     code:  assert Map.has_key?(state, :duplicates_filter) == false
     left:  true
     right: false
     stacktrace:
       test/pipelines/duplicates_filter_test.exs:31: (test)```

It's required to cleanup the Application state, as otherwise it will affect the second test. Please add the following setup:
```
  setup do
    on_exit(fn ->
      Application.put_env(:crawly, :item_id, nil)
    end)
  end
```

File: test/pipelines/csv_encoder_test.exs [hunk 2]
```diff
@@ -0,0 +1,19 @@
+defmodule Pipelines.CSVEncoderTest do
+  use ExUnit.Case, async: false
+
+  @valid %{first: "some", second: "data"}
+
+  test "CSVEncoder converts a single-level map to a csv string" do
+    Application.put_env(:crawly, :item, [:first, :second])
```
oltarasenko: Let's also add a cleanup, to revert the Application state after the test.

test/pipelines/validate_test.exs
```diff
@@ -0,0 +1,43 @@
...
+  @invalid_missing %{
+    title: "title"
+  }
+  @invalid_nil %{
+    title: "title",
+    author: nil
+  }
+
+  test "ValidateItem item with required fields are not dropped, item returned unchanged" do
+    Application.put_env(:crawly, :item, [:title, :author])
```
oltarasenko: Let's add a setup(teardown) methods here, to cleanup the application state

test/pipelines/validate_test.exs
```diff
@@ -0,0 +1,43 @@
...
+  test "ValidateItem items with missing required fields are dropped" do
+    Application.put_env(:crawly, :item, [:title, :author])
+    pipelines = [Crawly.Pipelines.Validate]
+    item = @invalid_missing
+    state = %{}
+
+    {false, _state} = Crawly.Utils.pipe(pipelines, item, state)
+  end
+
+  test "ValidateItem required item fields with nil are dropped" do
```
oltarasenko: What if we will rewrite test names in the following way (pattern):

`Drops items with required field set to nil`?

My idea is that, with this ValidateItem is already stated in the module name. Maybe lets not repeat ourselves. And another point, I like something like a command language. E.g. it explains what behaviour this test assures

oltarasenko:
I like the proposal in general. I would suggest some really minor tweaks before merging it!

test/pipelines/validate_test.exs
Ziinc: Yup I agree.

So test names should begin with a verb?

Ziinc: test names have been adjusted

PR #31, opened 2019-11-30
Pipe adjustments to handle tuple-based declarations for pipelines and middlewares

Ziinc:
todos:

- [x]  update pipelines and corresponding tests once #29 is merged
    - [x] WriteToFile pipeline remaining

- [x] update middlewares and corresponding tests

- [x] add custom middleware example

- [ ] logger warn users that are still using global env declaration, show example of correct tuple declation w/ example
- [x] update user docs, deprecation warning for global configs, quickstart
- [x] update dev docs, doc for `Crawly.Utils.pipe`, best prac for handling pipeline arguments

related to #22

### DIFF
--- a/lib/crawly/pipelines/csv_encoder.ex
+++ b/lib/crawly/pipelines/csv_encoder.ex
@@ -1,13 +1,33 @@
 defmodule Crawly.Pipelines.CSVEncoder do
   @moduledoc """
-  Encodes a given item (map) into CSV
+  Encodes a given item (map) into CSV. Does not flatten nested maps.
+  ### Options
+  If no fields are given, the item is dropped from the pipeline.
+  - `:fields`, required: The fields to extract out from the scraped item. Falls back to the global config `:item`.
+
+  ### Example Usage
+    iex> item = %{my: "first", other: "second", ignore: "this_field"}
+    iex> Crawly.Pipelines.CSVEncoder.run(item, %{}, fields: [:my, :other])
+    {"first,second", %{}}
   """
   @behaviour Crawly.Pipeline
+  require Logger
 
   @impl Crawly.Pipeline
-  def run(item, state) do
-    case Application.get_env(:crawly, :item) do
+  @spec run(map, map, fields: list(atom)) ::
+          {false, state :: map} | {csv_line :: String.t(), state :: map}
+  def run(item, state, opts \\ []) do
+    opts = Enum.into(opts, %{fields: nil})
+    fields = Map.get(opts, :fields) || Application.get_env(:crawly, :item)
+
+    case fields do
       :undefined ->
+        # only for when both tuple and global config is not provided
+
+        Logger.info(
+          "Dropping item: #{inspect(item)}. Reason: No fields declared for CSVEncoder"
+        )
+
         {false, state}
 
       fields ->
--- a/lib/crawly/utils.ex
+++ b/lib/crawly/utils.ex
@@ -35,19 +35,32 @@ defmodule Crawly.Utils do
   @doc """
   Pipeline/Middleware helper
 
-  Executes a given list of pipelines on the given item, mimics filtermap
-  behavior (but probably in a more complex way). Takes an item and state  and
-  passes it through a list of modules which implements a pipeline behavior,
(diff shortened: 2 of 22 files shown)


Ziinc:
codebase should avoid using Application.put_env and Application get_env. 

Since this PR also affects middleware declarations, the middlewares will also have to be updated.

`Application get_env` can be used as a fallback for getting args during transition period

Ziinc:
- [ ]  dev docs, guide on creating custom pipelines

Ziinc:
@oltarasenko Completed, could you review? 

The logger warnings can be moved to a separate issue, as they're not critical and documentation has already been updated.

lib/crawly/pipelines/csv_encoder.ex
```diff
@@ -1,13 +1,33 @@
...
+  require Logger
 
   @impl Crawly.Pipeline
-  def run(item, state) do
-    case Application.get_env(:crawly, :item) do
+  @spec run(map, map, fields: list(atom)) ::
+          {false, state :: map} | {csv_line :: String.t(), state :: map}
+  def run(item, state, opts \\ []) do
+    opts = Enum.into(opts, %{fields: nil})
+    fields = Map.get(opts, :fields) || Application.get_env(:crawly, :item)
```
oltarasenko: Why do we need this `Application.get_env(:crawly, :item)` fallback here?

lib/crawly/utils.ex
```diff
@@ -60,16 +73,30 @@ defmodule Crawly.Utils do
   def pipe(_, false, state), do: {false, state}
 
   def pipe([pipeline | pipelines], item, state) do
+    {module, args} =
```
oltarasenko: As I understand you're wrapping the Middlewares so they have a common look. {module, args}. I think it's better not to do it here. E.g. pipe should pipe. Not it wrapps and pipes :). Lets move the wrapping part away. Probably to the storage worker. Maybe here: https://github.com/oltarasenko/crawly/blob/ede6a6ebf39d63de79e8fc266bf80e07f1e79e39/lib/crawly/requests_storage/requests_storage_worker.ex#L65

oltarasenko:
I have some really minor comments I want to address, otherwise great job @Ziinc !

** I still feel a bit uncomfortable about the fact that we don't have an explicit item definition outside of the pipelines scope.

oltarasenko:
@Ziinc Please also sqash commits!

lib/crawly/pipelines/csv_encoder.ex
Ziinc: Because if the fields are not passed as a tuple config, the , `opts.fields` will be nil. 

To be backwards compatible, the fallback is needed for users that are not using tuple configs.

lib/crawly/utils.ex
Ziinc: I don't quite understand what you mean. Are you referring to the naming use of `module` instead of `pipeline`? For this, part, I had thought that module was appropriate since both the middlewares and pipelines use `Util.pipe` and should behave the same way.

If I were to move the tuple-format checks to the request storage worker, I would have to duplicate the code to the item storage worker as well.

Ziinc:
I think I can adjust the docs such that the tuple definitions act more as an escape hatch for advanced use cases. Would that be alright @oltarasenko

lib/crawly/utils.ex
oltarasenko: Oh.  I did not think about Item pipelines. I general I would prefer pipe not to be responsible for wrapping/unwrapping. But, you're right. We don't have options atm :(

oltarasenko:
Looks good to me. But still, I would want to ask you to rewrite the commit message & fix the:
```
warning: variable "value" is unused
  test/middlewares/user_agent_test.exs:20
Compilation failed due to warnings while using the --warnings-as-errors option
The command "mix coveralls.travis" exited with 1.
cache.2
store build cache
```

reported by Travis.

Ziinc:
Made the changes. Not so used to squashing commits. 😄

oltarasenko:
Great work! Thanks for your contribution. I have approved the PR!

PR #32, opened 2019-12-07
Migrate the static documentation to ex_doc

oltarasenko:


Ziinc:
Will we be using https://oltarasenko.github.io/crawly/ or the one on hexdocs https://hexdocs.pm/crawly/api-reference.html ? if we're using the hexdocs endpoint, then i'll remove files for the github pages

Ziinc:
the site at hexdocs has a higher google ranking

oltarasenko:
I would move to hexdocs eventually. The problem here is that current github pages has several links from articles I have written before. So I would want to keep it for some time (or to make a link from github pages to hexdocs (I kind of like what I see on hexdocs already))

Ziinc:
Looks good to me, I've made more edits and merged some sections of the guide. I've also grouped the modules together based on reader relevance.

I think the docs can be made more concise, but I'll do that on a separate PR.

Ziinc:
travis checks failed for a non-related test. Good to merge

oltarasenko:
I think one of the tests has a race condition. I will merge this now.

Issue #33, opened 2019-12-09
Add options for HTTPoison


srmor03:
How it is possible to implement in Crawly adding options for HTTPoison.
For example, such:

```

url = "https://example.com/api/endpoint_that_needs_a_bearer_token"
headers = []
**options = [ssl: [{:versions, [:'tlsv1.2']}], recv_timeout: 500]**
{:ok, response} = HTTPoison.get(url, headers, **options**)
```

And so that it can then be used in a spider.

Ziinc:
You want to do it on a per-request, per-spider, or global basis?

srmor03:
Global basis

oltarasenko:
I agree. I need to add it. I will schedule the change for the next release.

oltarasenko:
@srmor03 I have made the following change to address your issue: https://github.com/oltarasenko/crawly/pull/38. Hopefully, it can be reviewed and approved soon!

Ziinc:
The HTTPoison options for ssl can be set through a custom request middleware. Configuration and secrets can be passed through Application.get_env in your config.exs.

Please refer to the latest documentation in master:

https://github.com/oltarasenko/crawly/blob/master/documentation/basic_concepts.md#request-middleware-example---add-a-proxy

The latest documentation is not up yet on hex.pm,so refer to the markdown file for now. For v0.7.0 and below, tuple based middleware/pipeline declarations are not available.

Closing.

Issue #27, opened 2019-11-29
Write an example of Crawly with splash integrated as a proxy


oltarasenko:

Ziinc:
might need to consider that in a mutli-spider situation, some spiders do not need the browser rendering.  Maybe it can be on the spider implementation level, where browser rendering is declared with a flag (or something like that)

Ziinc:
related to #18

oltarasenko:
Yes, in general we need to be able to setup different HTTP clients for different spiders

Ziinc:
Thinking a bit more:
Could it be at a request level? What if a site's structure has certain parts that require JS, and certain parts that don't. It would be unnecessary and inefficient to use a browser for parts which don't require JS.

jallum:
I think that using the request makes the most sense.  The spider is in a good position to make the decision as to what kind of request should be made on it's behalf. 

Thinking out loud, what about returning plain url-strings in the `ParsedItem`, and then crawly could pass those url-strings back to a new (optional) callback on the spider to produce the request given the url?  At that point, the spider could configure a request and return it.  The default implementation of `build_request` (or whatever better name people can think of) would also serve to eliminate a bit of boilerplate that seems to find it's way into all my spiders:

```
 # Convert URLs into requests
    requests =
      urls
      |> Enum.map(&build_absolute_url/1)
      |> Enum.map(&Crawly.Utils.request_from_url/1)
```

This would also serve to eliminate the construction of unecessary request-stuff on the occasion that the request is a duplicate (or is dropped for some other reason.)  crawly would only call the spider's `build_request` on urls it actually intends to follow.

Ziinc:
@jallum I think improving `Crawly.Utils.request_from_url/1` could be split to another issue where relative/absolute url checks and building are done automatically.

Maybe a `:browser` boolean flag in `Crawly.Request`?

jallum:
Agreed.

oltarasenko:
It turns out that splash does not yet have a proxy interface :(. I don't really like this way of doing requests:
```curl 'http://localhost:8050/render.html?url=http://domain.com/page-with-javascript.html&timeout=10&wait=0.5'
```

So probably will switch back to the original idea of headless browsers

Ziinc:
But this would be all within a middleware, right? There

oltarasenko:
[...]

PR #34, opened 2019-12-10
Pluggable fetchers

Ziinc:
For #27 and #33

### DIFF
--- a/lib/crawly/fetchers/fetcher.ex
+++ b/lib/crawly/fetchers/fetcher.ex
@@ -2,12 +2,13 @@ defmodule Crawly.Fetchers.Fetcher do
   @moduledoc """
   A behavior module for defining Crawly Fetchers
 
-  A fetcher is expected to implement a fetch callback which should return a
-  Crawly.Response
+  A fetcher is expected to implement a fetch callback which should take
+  Crawly.Request, HTTP client options and return Crawly.Response.
   """
 
-  @callback fetch(request) :: {:ok, response} | {:error, reason}
+  @callback fetch(request, options) :: {:ok, response} | {:error, reason}
             when request: Crawly.Request.t(),
                  response: Crawly.Response.t(),
+                 options: map(),
                  reason: term()
 end
--- a/lib/crawly/fetchers/fetcher.ex
+++ b/lib/crawly/fetchers/fetcher.ex
@@ -1,0 +1,13 @@
+defmodule Crawly.Fetchers.Fetcher do
+  @moduledoc """
+  A behavior module for defining Crawly Fetchers
+
+  A fetcher is expected to implement a fetch callback which should return a
+  Crawly.Response
+  """
+
+  @callback fetch(request) :: {:ok, response} | {:error, reason}
+            when request: Crawly.Request.t(),
+                 response: Crawly.Response.t(),
+                 reason: term()
+end
--- a/lib/crawly/response.ex
+++ b/lib/crawly/response.ex
@@ -1,0 +1,19 @@
+defmodule Crawly.Response do
+  @moduledoc """
+  Define Crawly response structure
+  """
+
+  defstruct body: nil,
+            headers: [],
+            request: nil,
+            request_url: nil,
+            status_code: nil
+
+  @type t :: %__MODULE__{
+               body: term(),
+               headers: list(),
+               request: Crawly.Request.t(),
+               request_url: Crawly.Request.url(),
+               status_code: integer()
+             }
+end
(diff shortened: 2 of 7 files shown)


oltarasenko:
@Ziinc I have updated it to the implementation I have in mind, for now, we don't take into account any sort of per-spider configuration. However, I will address this issue separately.

oltarasenko:
@Ziinc any feedback here? I want to merge it, as I also have the second PR here, e.g. the Splash fetcher.

lib/crawly/fetchers/fetcher.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Crawly.Fetchers.Fetcher do
+  @moduledoc """
+  A behavior module for defining Crawly Fetchers
+
+  A fetcher is expected to implement a fetch callback which should return a
+  Crawly.Response
```
Ziinc: Perhaps add in that it shoukd receive a Crawly.Request (and return a Crawly. Response)

lib/crawly/response.ex
```diff
@@ -0,0 +1,19 @@
...
+            headers: [],
+            request: nil,
+            request_url: nil,
+            status_code: nil
+
+  @type t :: %__MODULE__{
+               body: term(),
+               headers: list(),
+               request: Crawly.Request.t(),
+               request_url: Crawly.Request.url(),
```
Ziinc: Is the request_url field necessary? It is already in the Request map. For convenience?

lib/crawly/worker.ex
```diff
@@ -66,7 +70,15 @@ defmodule Crawly.Worker do
              response: HTTPoison.Response.t(),
              result: {:ok, response, spider_name} | {:error, term()}
   defp get_response({request, spider_name}) do
-    case HTTPoison.get(request.url, request.headers, request.options) do
+    # check if spider-level fetcher is set. Overrides the globally configured fetcher.
+    # if not set, log warning for explicit config preferred, get the globally-configured fetcher. Defaults to FwtchWithHTTPoison
+    fetcher = Application.get_env(
```
Ziinc: Maybe we should allow tuple-style fetcher declarations too, to pass default options to the fetcher.

lib/crawly/fetchers/fetcher.ex
oltarasenko: Done!

lib/crawly/response.ex
oltarasenko: I think it's driven by HTTPoison. My understanding that request url might be different. For example in the case of redirect:
```
iex(1)> HTTPoison.get("http://meta.ua")
{:ok,
 %HTTPoison.Response{
   body: "<html>\r\n<head><title>301 Moved Permanently</title></head>\r\n<body bgcolor=\"white\">\r\n<center><h1>301 Moved Permanently</h1></center>\r\n<hr><center>nginx/1.14.0</center>\r\n</body>\r\n</html>\r\n",
   headers: [
     {"Server", "nginx/1.14.0"},
     {"Date", "Mon, 30 Dec 2019 13:00:44 GMT"},
     {"Content-Type", "text/html"},
     {"Content-Length", "185"},
     {"Connection", "keep-alive"},
     {"Location", "https://meta.ua/"}
   ],
   request: %HTTPoison.Request{
     body: "",
     headers: [],
     method: :get,
     options: [],
     params: %{},
     url: "http://meta.ua"
   },
   request_url: "http://meta.ua",
   status_code: 301
 }}```

File: lib/crawly/worker.ex (hunk 3 above)
Comment (oltarasenko) [reply lib/crawly/worker.ex]: +1

**Comment by oltarasenko** [comment]
@Ziinc I have addressed your comments in the most recent commit.

File: lib/crawly/response.ex (hunk 2 above)
Comment (Ziinc) [reply lib/crawly/response.ex]: I did some digging, turns out that HTTPoison mimic's python's request library, such that the `request_url` key reflects the final requested url made, after applying modifications such as params.

relevant HTTPoison issue:
https://github.com/edgurgel/httpoison/issues/270

relevant python library:
https://2.python-requests.org//en/master/user/quickstart/#passing-parameters-in-urls

I think it will be good to add in docs on what request_url is used for in the `Crawly.Response` struct, as it will also apply to how Fetchers should output the Response.

**Comment by Ziinc** [comment]
Oh, just thought of another point on the tuple-based config passing.

I think that for introducing beginner users, we should use the normal atom-based definitions, while the tuple-based configurations should be allowed for more advanced users only.

This is because, from a beginner point of view, the default options would be more than necessary to get up and running, while more advanced users would of course want to tweak the options. It would also provide unnecessary friction to learn about tuple-based options when all the beginner wants is to try it out.

So, for example, when declaring  the fetcher in the tutorials, we should use:
```elixir
config, :crawly,
    fetcher: Crawly.Fetchers.HTTPoisonFetcher
```
while also allowing advanced users to use:
```elixir
    fetcher: {Crawly.Fetchers.HTTPoisonFetcher, [my: options]
```
It will mean having to create a util function to standardize the tuple-based config unwrapping across the codebase.

For now, we can merge this PR first (since it blocks other issues), and create a separate issue for this

PR #38, opened 2019-12-20
Allow to specify HTTPoison options directly in config

oltarasenko:
Made in order to address the: https://github.com/oltarasenko/crawly/issues/33

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -47,6 +47,9 @@ config :crawly,
   # Stop spider if it does crawl fast enough
   closespider_timeout: 20,
   concurrent_requests_per_domain: 5,
+  httpoison_options: [
+    {:follow_redirects, true},
+  ],
   follow_redirects: true,
   # Request middlewares
   middlewares: [
--- a/config/test.exs
+++ b/config/test.exs
@@ -20,7 +20,10 @@ config :crawly,
   # Stop spider if it does crawl fast enough
   closespider_timeout: 20,
   concurrent_requests_per_domain: 5,
-  follow_redirects: true,
+  httpoison_options: [
+    {:follow_redirects, true},
+  ],
+
   # Request middlewares
   middlewares: [
     Crawly.Middlewares.DomainFilter,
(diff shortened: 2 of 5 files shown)


Ziinc:
From my understanding of `worker.ex`, the HTTPoison options are already getting passed to the `HTTPoison.get` through the `:options` key on the `Crawly.Request` struct

In `worker.ex:69`:
image


I don't think additional global configs are required. Just need additional documentation on how to create a custom middlewares to add in request options

Ziinc:
Additionally, it would be more appropriate to abstract out the `:proxy` and `:follow_redirects` global options to individual middlewares.

Also, I notice that the `follow_redirects` config key is being used or passed to HTTPoison at all. 

I'm going to be finishing up #31 today, could you use that as a basis for writing the middlewares for the two options (proxy and follow_redirects)? I will be adding in docs on #31 on how to write a custom middleware, so it won't be necessary to do so on this branch

Another note is that the proxy middleware needs to accept two options, the proxy host and port, and a proxy_auth key as well, as HTTPoison doesn't parse a proxy string in the format of `http://user:pass@host:port`.
I'll chip in once I finish #31

oltarasenko:
Hmm... Yeah @Ziinc , it's complicated I agree.

I kind of agree with your comments!
I think I can try using the middlewares for this purpose as well, however, I am not sure we should do 
it. E.g. I prefer to use middlewares for things which require more-or-less complicated ongoing operations on the requests. 

So to say from my point of view it probably makes sense using middlewares for the case of proxies (where we might want to direct requests via different sources). But maybe I would not want to use them to hardcode the ssl options.

What do you think?

Ziinc:
From my understanding of the `:ssl` options used by the erlang module (which HTTPoison uses), the developer will have to write a custom middleware anyway to be able to read/reference key files anyway. I have already updated the documentation under `Basic Concepts` in #31 with an example on how to write a custom middleware to add options to `Crawly.Request` (which are then passed to HTTPoison`)

I don't think merging this PR is necessary, just maybe need to add even more to the docs?

Ziinc:
I think the points raised in :
- https://github.com/oltarasenko/crawly/pull/38#issuecomment-567874894
- https://github.com/oltarasenko/crawly/pull/38#pullrequestreview-335180454

would be better to tackle, as this PR is not necessary since options can already be set through the request.
@oltarasenko close if you agree, thanks

Ziinc:
Closing as documentation on how to pass HTTPoison request options has been added into documentation in #31

PR #39, opened 2019-12-20
Add retries support

oltarasenko:
Try to fetch unsuccessful requests more then once. The amount
of attempts is controlled by the max retries setting.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -33,7 +33,7 @@ config :crawly, Crawly.Worker, client: HTTPoison
 
 config :crawly,
   fetcher: {Crawly.Fetchers.HTTPoisonFetcher, []},
-
+  max_retries: 3,
   # User agents which are going to be used with requests
   user_agents: [
     "Mozilla/5.0 (Macintosh; Intel Mac OS X x.y; rv:42.0) Gecko/20100101 Firefox/42.0",
--- a/config/config.exs
+++ b/config/config.exs
@@ -33,7 +33,9 @@ config :crawly, Crawly.Worker, client: HTTPoison
 
 config :crawly,
   fetcher: {Crawly.Fetchers.HTTPoisonFetcher, []},
-  max_retries: 3,
+  retry: [
+    {:max_retries, 3},
+    {:ignored_middlewares, [Crawly.Middlewares.UniqueRequest]}],
   # User agents which are going to be used with requests
   user_agents: [
     "Mozilla/5.0 (Macintosh; Intel Mac OS X x.y; rv:42.0) Gecko/20100101 Firefox/42.0",
@@ -63,8 +65,9 @@ config :crawly,
     Crawly.Pipelines.JSONEncoder
   ]
 
-config :crawly, Crawly.Pipelines.WriteToFile,
-  folder: "/tmp",
-  extension: "jl"
+config :crawly,
+       Crawly.Pipelines.WriteToFile,
+       folder: "/tmp",
+       extension: "jl"
 
- import_config "#{Mix.env}.exs"
+import_config "#{Mix.env}.exs"
--- a/lib/crawly/middlewares/robotstxt.ex
+++ b/lib/crawly/middlewares/robotstxt.ex
@@ -21,9 +21,10 @@ defmodule Crawly.Middlewares.RobotsTxt do
   require Logger
 
   def run(request, state, _opts \\ []) do
-    case Gollum.crawlable?("Crawly", request.url) do
+    url = Crawly.Request.url(request)
+    case Gollum.crawlable?("Crawly", url) do
       :uncrawlable ->
-        Logger.debug("Dropping request: #{request.url} (robots.txt filter)")
+        Logger.debug("Dropping request: #{url} (robots.txt filter)")
 
         {false, state}
 
--- a/lib/crawly/middlewares/unique_request.ex
+++ b/lib/crawly/middlewares/unique_request.ex
@@ -5,19 +5,15 @@ defmodule Crawly.Middlewares.UniqueRequest do
   """
   require Logger
 
-  # Allow to re-schedule a request if retries are required
-  def run(%Crawly.Request{retries: retries} = request, state) when retries > 0 do
-    {request, state}
-  end
-
   def run(request, state) do
     unique_request_seen_requests =
       Map.get(state, :unique_request_seen_requests, %{})
 
-    case Map.get(unique_request_seen_requests, request.url) do
+    url = Crawly.Request.url(request)
+    case Map.get(unique_request_seen_requests, url) do
       nil ->
         unique_request_seen_requests =
-          Map.put(unique_request_seen_requests, request.url, true)
+          Map.put(unique_request_seen_requests, url, true)
 
         new_state =
           Map.put(
@@ -30,7 +26,7 @@ defmodule Crawly.Middlewares.UniqueRequest do
 
       _ ->
         Logger.debug(
-          "Dropping request: #{request.url}, as it's already processed"
+          "Dropping request: #{url}, as it's already processed"
         )
 
         {false, state}
--- a/lib/crawly/middlewares/unique_request.ex
+++ b/lib/crawly/middlewares/unique_request.ex
@@ -1,8 +1,14 @@
 defmodule Crawly.Middlewares.UniqueRequest do
   @moduledoc """
-  Avoid scheduling requests for the same pages.
+  Avoid scheduling requests for the same pages. However if retry is requested
+  the request is still allowed.
   """
   require Logger
+
+  # Allow to re-schedule a request if retries are required
+  def run(%Crawly.Request{retries: retries} = request, state) when retries > 0 do
+    {request, state}
+  end
 
   def run(request, state) do
     unique_request_seen_requests =
@@ -30,4 +36,5 @@ defmodule Crawly.Middlewares.UniqueRequest do
         {false, state}
     end
   end
+
 end
--- a/lib/crawly/middlewares/user_agent.ex
+++ b/lib/crawly/middlewares/user_agent.ex
@@ -18,19 +18,25 @@ defmodule Crawly.Middlewares.UserAgent do
   """
   require Logger
 
-  def run(request, state, opts \\ []) do
+  def run(request, state, opts \\ %{}) do
     opts = Enum.into(opts, %{user_agents: nil})
+    headers = Crawly.Request.headers(request)
 
-    new_headers = List.keydelete(request.headers, "User-Agent", 0)
-
+    # Getting user agent from a list defined by a middleware
     user_agents =
       Map.get(opts, :user_agents) ||
         Application.get_env(:crawly, :user_agents, ["Crawly Bot 1.0"])
+    ua = {"User-Agent", Enum.random(user_agents)}
 
-    useragent = Enum.random(user_agents)
+    new_headers =
+      case List.keyfind(headers, "User-Agent", 0, nil) do
+        nil ->
+          [ua | headers]
+        _ ->
+          List.keyreplace(headers, "User-Agent", 0, {"User-Agent", ua})
+      end
 
-    new_request =
-      Map.put(request, :headers, [{"User-Agent", useragent} | new_headers])
+    new_request = Crawly.Request.headers(request, new_headers)
 
     {new_request, state}
   end
--- a/lib/crawly/request.ex
+++ b/lib/crawly/request.ex
@@ -23,13 +23,13 @@ defmodule Crawly.Request do
   @type retries :: non_neg_integer()
   @type option :: {atom(), binary()}
   @opaque t :: %__MODULE__{
-               url: binary(),
-               headers: [header()],
-               prev_response: %{},
-               retries: 0,
-               middlewares: [],
-               options: [option()]
-             }
+                 url: binary(),
+                 headers: [header()],
+                 prev_response: %{},
+                 retries: 0,
+                 middlewares: [],
+                 options: [option()]
+               }
   ###===========================================================================
   ### API functions
   ###===========================================================================
@@ -41,8 +41,39 @@ defmodule Crawly.Request do
              headers: [term()],
              options: [term()],
              request: Crawly.Request.t()
+
   def new(url, headers \\ [], options \\ []) do
-    %Crawly.Request{url: url, headers: headers, options: options}
+    # Define a list of middlewares which are used by default to process
+    # incoming requests
+    default_middlewares = [
+      Crawly.Middlewares.DomainFilter,
+      Crawly.Middlewares.UniqueRequest,
+      Crawly.Middlewares.RobotsTxt
+    ]
+
+    middlewares =
+      Application.get_env(:crawly, :middlewares, default_middlewares)
+
+    new(url, headers, options, middlewares)
+  end
+
+  @doc """
+  Same as Crawly.Request.new/3 from but allows to specify middlewares as the 4th
+  parameter.
+  """
+  @spec new(url, headers, options, middlewares) :: request
+        when url: binary(),
+             headers: [term()],
+             options: [term()],
+             middlewares: [term()], # TODO: improve typespec here
+             request: Crawly.Request.t()
+  def new(url, headers, options, middlewares) do
+    %Crawly.Request{
+      url: url,
+      headers: headers,
+      options: options,
+      middlewares: middlewares
+    }
   end
 
   @doc """
@@ -115,14 +146,14 @@ defmodule Crawly.Request do
              options_field: binary()
   def options(request, new_options),
       do: %Crawly.Request{request | options: new_options}
-      
+
   @doc """
   Access retries field from Crawly.Request
   """
   @spec retries(request) :: retries_field
         when request: Crawly.Request.t(),
              retries_field: [term()]
-  def retries(request), do: request.headers
+  def retries(request), do: request.retries
 
   @doc """
   Set retries field in Crawly.Request
--- a/lib/crawly/request.ex
+++ b/lib/crawly/request.ex
@@ -4,7 +4,7 @@ defmodule Crawly.Request do
 
   Defines Crawly request structure.
   """
-  defstruct url: nil, headers: [], prev_response: nil, options: []
+  defstruct url: nil, headers: [], prev_response: nil, options: [], retries: 0
 
   @type header() :: {key(), value()}
   @type url() :: binary()
@@ -12,12 +12,15 @@ defmodule Crawly.Request do
   @typep key :: binary()
   @typep value :: binary()
 
+  @type retries :: non_neg_integer()
+
   @type option :: {atom(), binary()}
 
   @type t :: %__MODULE__{
     url: url(),
     headers: [header()],
     prev_response: %{},
+    retries: 0,
     options: [option()]
   }
 end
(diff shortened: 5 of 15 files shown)


test/worker_test.exs
```diff
@@ -130,15 +138,70 @@ defmodule WorkerTest do
          }}
       end)
 
-      Process.sleep(1_000)
+      send(context.crawler, :work)
       assert Process.alive?(context.crawler)
     end
+
+    test "Requests are dropped after 3 attempts", context do
```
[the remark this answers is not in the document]
Ziinc: I'm thinking that there are two options for the worker to know if a request should be sent back for retry:
1. A request key boolean flag (like`retry_if_fail`) that any middleware can set to true, and the worker simply checks if it is true or false
2. The worker checks if a retry middleware is declared. If so, automatically retries failed requests.

I think option 1 is much better, since it isn't rigid, and gives more flexibility in whether a retry should be done and it is on a per-request basis

lib/crawly/middlewares/unique_request.ex
```diff
@@ -1,9 +1,15 @@
 defmodule Crawly.Middlewares.UniqueRequest do
   @moduledoc """
-  Avoid scheduling requests for the same pages.
+  Avoid scheduling requests for the same pages. However if retry is requested
+  the request is still allowed.
   """
   require Logger
 
+  # Allow to re-schedule a request if retries are required
+  def run(%Crawly.Request{retries: retries} = request, state) when retries > 0 do
```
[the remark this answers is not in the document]
oltarasenko: The problem is that retries is introducing same requests several times. E.g. if we retry something it would mean that it has a stored fingerprint already. So we need to do something like a forced `store`. Otherwise, the retried requests can't really be unique.

lib/crawly/worker.ex
```diff
@@ -17,47 +17,44 @@ defmodule Crawly.Worker do
...
+
+      request ->
+        # Process the request using following group of functions
+        functions = [
+          {:get_response, &get_response/1},
+          {:parse_item, &parse_item/1},
+          {:process_parsed_item, &process_parsed_item/1}
+        ]
+
+        case :epipe.run(functions, {request, spider_name}) do
```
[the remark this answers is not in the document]
oltarasenko: I have chosen to use :epipe at the early stages for the product as it gives more control over the execution flow. 

I think the most appropriate alternative in Elixir is `with` operator. However I decided to use :epipe instead, as I don't like how with handles the multiple failing conditions. 

In any case this is not a part of given PR.

oltarasenko:
Hey, @Ziinc one high-level comment. The retries logic needs to be applied in a worker. Middleware can't really re-schedule requests. (Well at least if we don't introduce responses to middlewares).

E.g. we're making a decision if we plan to re-try requests after all middlewares are already applied.

config/config.exs
```diff
@@ -32,6 +32,7 @@ use Mix.Config
 config :crawly, Crawly.Worker, client: HTTPoison
 
 config :crawly,
+  max_retries: 3,
```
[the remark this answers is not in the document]
oltarasenko: Can we discuss the problem with global configs again, please? I am kind of missing the key problem there. 

E.g. do we have practical concerns when several applications would use different versions of Crawly as a dependency?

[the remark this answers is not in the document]
Ziinc: The issues with global configs that I can think of are:
1. a single application can be scraping multiple sites for different information, each might have retry requirements or different user agents required, so it makes sense to pass options directly to each pipeline such that the configs don't interfere with another pipeline.
2. Reduces reusability. For example, if I were to create a custom pipeline module that can apply a set of pipelines to an item if it meets a certain condition, for example, i can't reuse the `Pipelines.Validate` module because takes its config from the global config. In the case of retries, say i only want to apply retries to a particular type of scraped item, or if the url matches a certain condition, I can't simply create a pipeline that does a conditional if that then applies the pipeline.
For example, with direct options being passed, I can do this quite easily
```elixir
def MyPipelineExtension do
    def run( %{nested_item: nested } = item, state, opts) do
        Crawly.Utils.pipe([
            {Crawly.Pipeline.Validate, fields: [:my, :required, :fields] }
        ], item, state)
    end
    def run(item, state, _opts), do: {item, state}
end
```
It makes the built in modules so much more useful, and they act more like building blocks that can be applied according to the developer's needs. Using a global config is more like saying that the building block can only be done in one particular way, but the developer may actually need to adapt it for a whole different purpose

That's why i don't like global configs, because they make the built in modules essentially useless to me because I can't extend them like add special conditions or side effects.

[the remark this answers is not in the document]
Ziinc: The example is a simplification, of course, but the idea is that it allows more flexibility in declarations. So for example, if in the future we implement spider-level pipeline/middleware overrides, it would be just a simple pipeline that is customized for the site, instead of having to put_env and get_env and all that, which would change every pipeline's behaviour in the application.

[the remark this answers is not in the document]
oltarasenko: > a single application can be scraping multiple sites for different information, each might have retry requirements or different user agents required, so it makes sense to pass options directly to each pipeline such that the configs don't interfere with another pipeline.

That part is clear. I am thinking of adding the per-spider configs ASAP. (We had it in Scrapy, and I will add it here as well!)

> Reduces reusability. For example, if I were to create a custom pipeline module that can apply a set of pipelines to an item if it meets a certain condition, for example, i can't reuse the Pipelines. Validate module because takes its config from the global config. In the case of retries, say i only want to apply retries to a particular type of scraped item, or if the url matches a certain condition, I can't simply create a pipeline that does a conditional if that then applies the pipeline.
For example, with direct options being passed, I can do this quite easily ....

Yes, at some extent it's the case. But I would not expect Crawly to be used as a dependency somewhere, where some other app has crawly as a dependency as well.
Let's maybe keep some balance here. I agree that with middlewares/pipelines it's quite nice, as it makes the config cleaner.

oltarasenko:
@Ziinc Can we agree to support the global configs on the high level for retries? I will introduce the spider level configs, so they will take advantage over the global configs eventually

Ziinc:
Yup I understand that part about the rescheduling
https://github.com/oltarasenko/crawly/pull/39#discussion_r360400575

As for the global configs, how I handled it in #31 was to use it as a fallback, so if it wasn't provided in a tuple, it would check with a `get_env`

lib/crawly/middlewares/unique_request.ex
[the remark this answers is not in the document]
Ziinc: i see, that makes sense.Maybe the middleware should be renamed to convey its correct behaviour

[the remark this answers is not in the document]
oltarasenko: Yes, maybe. Any proposals? 

Also, I am not quite sure about this part:  `when retries > 0` e.g. it's not explicit ( E.g. it's not clear what is retry > 0 ). 

Should I add another field to a request, e.g. something like: is_retry_request: false/true, so it will be explicit?

config/config.exs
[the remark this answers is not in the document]
Ziinc: Yup of course, i understand what you mean about clean configs. But perhaps an escape hatch would be good for more advanced use cases, which is where support for the tuple-based would be good (so that new users can use the global configs, and when they need can switch over)

lib/crawly/middlewares/unique_request.ex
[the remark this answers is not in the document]
Ziinc: Coming up with a name is very hard because the two concerns are quite different. It's as if we're trying to fit a if-this-do-that into the name.

As suggested:
> A request key boolean flag (like `retry_if_fail`) that any middleware can set to true, and the worker simply checks if it is true or false.
Could something like this work? then the `UniqueRequest` middleware can just check if the request has this boolean flag and allow it to pass if so. A separate `Retry` middleware can then do the retry count increment and dropping

oltarasenko:
To summarize:

As I understand the main issue left for now is the fact that I had to "tune" the unique request middleware. Honestly speaking I think it's a bit of a problem, and I would want to keep middlewares untouched.

I have an idea in mind, which sounds quite suitable. E.g. I want to move middlewares to a request. So request will know how exactly it should be processed upon creation. E.g. there will be something like a route of processors to execute on the request.

I will try to submit a new version to a review as soon as possible. But it would be nice to do it after the #31 is merged.

Ziinc:
So, conceptually, the flow would be like this (taking into account #34 )?

image

oltarasenko:
I would describe it this way:
Crawly board

Ziinc:
Could it be something similar to this:
image

ideas:
1. unique request middleware remains unchanged
2. only new requests created through `new_request` pass through middlewares again
3. requests that are passed to `retry_request` do not go through middlewares, and instead go to the response storage
4. the Processors (pipeline behaviour modules) can call `new_request` and `retry_request`. Anything that falls under the concern of "converting a request into response" should be considered a Processor.
5. the fetcher processor  (for #34, maybe we can come up with a better name) converts the request to a response.
6. the retry processor receives the failed response from the previous pipeline module, and records the number of times it sees the particular failed request. Drops the request after count exceeds.


pros:
- reuse of the pipeline concept
- middlewares are untouched by retrying logic
- retry logic can be completely customized by advanced users.
- fetchers are pluggable, sole purpose is to convert a request into a response
- allows extending of the way `get_request` is called, allows for Processors to implement features like throttling. We can also abstract the backoff logic into a processor module as well.

cons: 
- new terminology, more customizability
- needs a diagram to be explained
- retry logic skips middlewares (though i think this is more of a pro than con, actually. Depends on scenario)

Ziinc:
@oltarasenko what are your thoughts on this? I can come up with a draft PR for the processors concept if you think that it is viable to solve both retries and pluggable response fetching.

oltarasenko:
@Ziinc I am not quite sure about making it more complex. I have another idea in mind, already working on a PR which should make the procedure simple and middlewares will not be changed. Just give me a bit of time to prepare it.

oltarasenko:
@Ziinc Please check the latest implementation here. The idea I have in mind is the following. 
1. We abstract the Crawly.Request even more (so it goes far away from the original HTTPoison.Request)
2. We add middlewares to the request structure. So the idea that request itself is something like a router, which identifies how to Route/Process itself.
3. It allows to "mangle" the request during retries, saying ok, when retrying I should ignore following middlewares: ...
4. It allows specifying middlewares on the spider level as well. (And on the parsing level if needed)

oltarasenko:
@Ziinc could you please have a glance?

lib/crawly/request.ex
```diff
@@ -2,22 +2,184 @@ defmodule Crawly.Request do
...
+  """
+  @spec url(request) :: url_field
+        when request: Crawly.Request.t(),
+             url_field: binary()
+  def url(request), do: request.url
+
+  @doc """
+  Set url field in Crawly.Request
+  """
+  @spec url(request, new_url) :: url_field
```
Ziinc: This function seems unnecessary, the request struct can be adjusted by the user directly.

lib/crawly/request.ex
```diff
@@ -2,22 +2,184 @@ defmodule Crawly.Request do
...
+        when request: Crawly.Request.t(),
+             new_url: binary(),
+             url_field: binary()
+  def url(request, new_url), do: %Crawly.Request{request | url: new_url}
+
+  @doc """
+  Access headers field from Crawly.Request
+  """
+  @spec headers(request) :: headers_field
```
Ziinc: Same as above, unnecessary function

lib/crawly/request.ex
```diff
@@ -2,22 +2,184 @@ defmodule Crawly.Request do
...
-  @type t :: %__MODULE__{
-    url: url(),
-    headers: [header()],
-    prev_response: %{},
-    options: [option()]
-  }
+  @doc """
+  Access prev_response field from Crawly.Request
+  """
+  @spec prev_response(request) :: prev_response_field
```
Ziinc: Same as above, can already be accessed through dot notation

lib/crawly/request.ex
```diff
@@ -2,22 +2,184 @@ defmodule Crawly.Request do
...
+  """
+  @spec options(request) :: options_field
+        when request: Crawly.Request.t(),
+             options_field: [term()]
+  def options(request), do: request.options
+
+  @doc """
+  Set options field in Crawly.Request
+  """
+  @spec options(request, new_options) :: options_field
```
Ziinc: Same as above, unnecessary fn

lib/crawly/request.ex
```diff
@@ -2,22 +2,184 @@ defmodule Crawly.Request do
...
+        when request: Crawly.Request.t(),
+             new_options: binary(),
+             options_field: binary()
+  def options(request, new_options),
+      do: %Crawly.Request{request | options: new_options}
+
+  @doc """
+  Access retries field from Crawly.Request
+  """
+  @spec retries(request) :: retries_field
```
Ziinc: Same as above.

lib/crawly/request.ex
```diff
@@ -2,22 +2,184 @@ defmodule Crawly.Request do
...
+        when request: Crawly.Request.t(),
+             new_retries: binary(),
+             retries_field: binary()
+  def retries(request, new_retries),
+      do: %Crawly.Request{request | retries: new_retries}
+
+  @doc """
+  Access middlewares field from Crawly.Request
+  """
+  @spec middlewares(request) :: middlewares_field
```
Ziinc: Same as above.

lib/crawly/middlewares/robotstxt.ex
```diff
@@ -21,9 +21,10 @@ defmodule Crawly.Middlewares.RobotsTxt do
   require Logger
 
   def run(request, state, _opts \\ []) do
-    case Gollum.crawlable?("Crawly", request.url) do
```
Ziinc: Unnecessary changes

lib/crawly/middlewares/user_agent.ex
```diff
@@ -18,19 +18,25 @@ defmodule Crawly.Middlewares.UserAgent do
   """
   require Logger
 
-  def run(request, state, opts \\ []) do
+  def run(request, state, opts \\ %{}) do
```
Ziinc: Elixir convention is for the default to be a keyword list

lib/crawly/requests_storage/requests_storage_worker.ex
```diff
@@ -63,16 +65,7 @@ defmodule Crawly.RequestsStorage.Worker do
...
-    # requests
-    default_middlewares = [
-      Crawly.Middlewares.DomainFilter,
-      Crawly.Middlewares.UniqueRequest,
-      Crawly.Middlewares.RobotsTxt
-    ]
-
-    middlewares =
-      Application.get_env(:crawly, :middlewares, default_middlewares)
+    middlewares = Crawly.Request.middlewares(request)
```
Ziinc: Can be accessed through dot notation instead

lib/crawly/worker.ex
```diff
@@ -80,11 +76,12 @@ defmodule Crawly.Worker do
     )
 
     case fetcher.fetch(request, options) do
-      {:ok, response} ->
+      {:ok, %HTTPoison.Response{status_code: 200} = response} ->
```
Ziinc: Should merge in latest master branch for changes on fetchers

oltarasenko:
Hey, @Ziinc regarding the:

> I don't see why we need to duplicate Elixir behaviour with the convenience functions on the Request module

I am quite open for a discussion and changes here. So lets have this conversation, let me try to explain what I had in mind, and let's make a decision based on the discussion.

Let me start with the idea:
This PR changes the Crawly.Request type definition to opaque type and the idea I am trying to push is that: now one should really reason what is the internal representation of Crawly.Request. Instead, you're working with the Request as with a black box through the getter and setter API. This is for example similar to https://hexdocs.pm/elixir/List.html. E.g. we don't know what is `list`, we just operate with it.

Fundamentally this approach has some pros and cons:
Pros:
1) It's possible to change the internal representation of the module without updating the entire codebase (as well as API stays the same);
2) Introduces something like separation of concerns. (This is probably driven by my bad experience with deeply nested maps, but in my practice, I saw cases when people was defining maps in other application, was sending them outside and pattern matched them in their own maps. Which was especially painful when the maps was very deep (e.g. had 3-4 layers of deepness)). So getting back to the topic: 
The fact that Crawly.Request is a map that is a coincidence (from the architecture point of view). If we would allow the outside world do something like request.url or %{request: req} then we will not really be able to encapsulate the Request data type :( 
Cons:
1) More code to write
2) Maybe harder to start using. E.g. you need to use functional API instead of built-in map notation.

Now finally:
In Crawly itself: I had to add default middlewares to request middlewares. This job can be done kind of seamlessly on a setter. And I did it via setter.

lib/crawly/request.ex
oltarasenko: @Ziinc lets address all concerns to API based requests altogether in the main idea explanation comment

lib/crawly/middlewares/user_agent.ex
oltarasenko: @Ziinc I am addressing these lines: https://github.com/oltarasenko/crawly/blob/master/lib/crawly/middlewares/user_agent.ex#L26-L28

E.g. it has `Map.get(opts, :user_agents)` so opts has to be a map. But now after looking on the pipeline behavior definition I think you're right. I need to fix it in a different way. will fix

lib/crawly/worker.ex
oltarasenko: I am not quite getting it, but otherwise, I think this is a result of rebase with latest master.

Ziinc:
OK, I understand what you want to achieve.
As for nested map pattern matching, I think that since pattern matching is an essential part of elixir, it doesn't make sense to completely prevent/block users from working around what is already familiar to them. For example, `Ecto.Changeset` is essentially a map that allows for pattern matching, although there are separate setter functions for updating the changeset, which may have some side effects.

Would having an opaque typespec make it harder to understand, especially with the internal implementation and data type hidden away?
Besides, when debugging, the first thing that would be done would be to call `IO.inspect/1`. The user would then realize that the underlying implementation is a map and then "hack together" a solution to update the map directly, bypassing the getters/setters.

I don't quite understand how it will introduce a separation of concerns. The `Request` struct represents how the fetcher will execute the request. All that it is concerned about is to represent the request. It doesn't need to know how or why it is updated to a certain state. If anything, by adding in getter/setter functions, introduces less separation by mixing the representation with the updating.

Another concern I have is the amount of code that would be required to implement the getters/setters that would result in much more maintenance. Considering the number of options and fields in [`HTTPoison.Request`](https://hexdocs.pm/httpoison/HTTPoison.Request.html#content), there would be at least an addition of a few thousand lines of code.

Ziinc:
I do think that introducing getter/setter functions would have minimal benefits compared to the costs (maintenance/complexity/difficulty to understand/duplicated efforts)

lib/crawly/worker.ex
Ziinc: because the latest changes in master allows for defining of a custom fetcher, which may not be a `HTTPoson.Response`. Maybe can change it to pattern match on just the map's first-level field `status_code`, or to pattern match on `%Crawly.Response{status_code: 200}`

oltarasenko:
Ok, I accept your points. I would re-raise this discussion separately on the next stages of the project development, once (or if) it will be more important for us to relay on opaque structures. For now I am closing this PR. Will re-do the re-tries part separately.

PR #44, opened 2020-01-02
fix square bracket / curly brace ordering

harlantwood:
PS: thanks for the nice tool!


oltarasenko:
Thanks for your contribution! @harlantwood !

PR #55, opened 2020-02-01
Add the option to change the port on application.ex

ogabriel:
Hi.

I was getting this error when trying to start two terminals at the same time:
```
** (Mix) Could not start application crawly: Crawly.Application.start(:normal, []) returned an error: shutdown: failed to start child: {:ranch_listener_sup, Crawly.API.Router.HTTP}
    ** (EXIT) shutdown: failed to start child: :ranch_acceptors_sup
        ** (EXIT) {:listen_error, Crawly.API.Router.HTTP, :eaddrinuse}
~
```

Because of that i searched for a option to change the port that cowboy uses, but i did't found one, so I have came up with this.
If you agree with the proposed changes, i will update all the documentation markdowns :+1: 

Great lib, i really helping me out:

### DIFF
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -153,7 +153,12 @@ Example:
 
 ### fetcher :: atom()
 
-default: Crawly.Fetchers.HTTPoisonFetcher 
+default: Crawly.Fetchers.HTTPoisonFetcher
 
-Allows to specify a custom HTTP client which will be performing request to the crawl
-target.
+Allows to specify a custom HTTP client which will be performing request to the crawler target.
+
+### port :: pos_integer()
+
+default: 4001
+
+Allows to specify a custom port to start the application. That is important when running more than one application in a single machine, in which case shall not use the same port as the others.
--- a/README.md
+++ b/README.md
@@ -79,7 +79,8 @@ historical archival.
        {Crawly.Pipelines.DuplicatesFilter, item_id: :title},
        Crawly.Pipelines.JSONEncoder,
        {Crawly.Pipelines.WriteToFile, extension: "jl", folder: "/tmp"} # NEW IN 0.7.0
-      ]
+      ],
+      port: 4001
    ```
 5. Start the Crawl:
    - `$ iex -S mix`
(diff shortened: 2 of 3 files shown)


oltarasenko:
Looks good from my point of view. If you could update the doc as well I will get it merged!

Thanks for the hint 👍

oltarasenko:
@ogabriel

ogabriel:
Ok, since I got your approval, I will add the documentation to this option

When I update the documentation, this will go automatically to hexdocs when a new version is released, or I have to update something else other than this project?

oltarasenko:
@ogabriel The documentation is published automatically upon the release!

documentation/configuration.md
```diff
@@ -153,7 +153,12 @@ Example:
...
-Allows to specify a custom HTTP client which will be performing request to the crawl
-target.
\ No newline at end of file
+Allows to specify a custom HTTP client which will be performing request to the crawler target.
+
+### port :: pos_integer()
+
+default: 4001
+
+Allows to specify a custom port to start the application, that is important when running more than one crawler (or other applications) in a single machine.
```
oltarasenko: Just to double-check. It's possible to run multiple Crawlers. You can't start multiple instances of Crawly. And in general, it's possible to run as many individual Crawlers as you want. could you please update the text accordingly, and I will gladly merge it!

ogabriel: Ok, it really don't make sense.
What do you think of:
```
Allows to specify a custom port to start the application, that is important when running more than one applications that can use the same port  in a single machine.
```

oltarasenko: Sounds good

ogabriel: I changed it a little bit, I it is clearer in this way

Ziinc:
@oltarasenko the commits can be squashed on merge by github, it's an option in the drop-down

oltarasenko:
@Ziinc yes, but I am not sure if it will not rewrite the history (I want @ogabriel to be the author of the commit). Don't want to check this way...

PR #56, opened 2020-02-15
Add request options middleware

oltarasenko:
The request options middleware is now used as a centralized place
to add request options.

### DIFF
--- a/documentation/basic_concepts.md
+++ b/documentation/basic_concepts.md
@@ -113,6 +113,13 @@ Built-in middlewares:
 2. `Crawly.Middlewares.RobotsTxt` - this middleware ensures that Crawly respects the robots.txt defined by the target website.
 3. `Crawly.Middlewares.UniqueRequest` - this middleware ensures that crawly would not schedule the same URL(request) multiple times.
 4. `Crawly.Middlewares.UserAgent` - this middleware is used to set a User Agent HTTP header. Allows to rotate UserAgents, if the last one is defined as a list.
+5. `Crawly.Middlewares.RequestOptions` - allows to set additional request options, for example timeout, of proxy string (at this moment the options should match options of the individual fetcher (e.g. HTTPoison))
+   
+   
+   Example:
+   ```elixir
+    {Crawly.Middlewares.RequestOptions, [timeout: 30_000, recv_timeout: 15000]}
+   ```
 
 ### Item Pipelines
 
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -8,7 +8,7 @@ A basic example:
 config :crawly,
   pipelines: [
     # my pipelines
-  ]
+  ],
   middlewares: [
     # my middlewares
   ]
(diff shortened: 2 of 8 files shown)


test/middlewares/request_options_test.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Middlewares.OptionsTest do
```
Ziinc: Should follow file name and the tested module's name , so RequestOptionsTest

lib/crawly/request.ex
```diff
@@ -48,6 +48,7 @@ defmodule Crawly.Request do
     # incoming requests
     default_middlewares = [
       Crawly.Middlewares.DomainFilter,
```
Ziinc: Not necessary, since it defaults to an empty list in the request struct anyway

documentation/basic_concepts.md
```diff
@@ -113,6 +113,13 @@ Built-in middlewares:
 2. `Crawly.Middlewares.RobotsTxt` - this middleware ensures that Crawly respects the robots.txt defined by the target website.
 3. `Crawly.Middlewares.UniqueRequest` - this middleware ensures that crawly would not schedule the same URL(request) multiple times.
 4. `Crawly.Middlewares.UserAgent` - this middleware is used to set a User Agent HTTP header. Allows to rotate UserAgents, if the last one is defined as a list.
+5. `Crawly.Middlewares.RequestOptions` - allows to set additional request options, for example timeout, of proxy string.
+   
+
```
Ziinc: Should we add in a blurb that says that the options handling behaviour is handled by indivudls fetcher modules? For e.g. For a HTTPoison module, the options should match HTTPoison's request options

documentation/configuration.md
```diff
@@ -79,14 +79,15 @@ config :crawly,
 
 ### middlewares :: [module()]
 
-```elixir
 The default middlewares are as follows:
```
Ziinc: Should be "an example of middlewares", since it does not actually match the defaults used in worker

Ziinc:
Logic looks good. Though technically, if all options that are for the fetcher are all handled by this one module, then we could technically merge the UserAgent module into this. Just some food for thought

test/middlewares/request_options_test.exs
```diff
@@ -0,0 +1,18 @@
...
+
+  test "Options are added to request settings" do
+    req = Crawly.Request.new("http://example.com")
+    middlewares = [
+      {
+        Crawly.Middlewares.RequestOptions,
+        [timeout: 30_000, recv_timeout: 15000]
+      }
+    ]
+    state = %{spider_name: :test_spider}
```
Ziinc: Is the spider_name key relevant for the test? Could just use an empty map if there are no relevant state assertion

test/middlewares/request_options_test.exs
oltarasenko: 👍

lib/crawly/request.ex
oltarasenko: yeah, agreed

documentation/basic_concepts.md
oltarasenko: yes, make sense

documentation/configuration.md
oltarasenko: agreed

test/middlewares/request_options_test.exs
oltarasenko: yes, we don't need any special state as I see

oltarasenko:
I am not quite sure about headers. I would treat them separately. E.g. I want to separate options from headers. 

Also one of the things to address in the future - things set on spider level should override middleware based settings.

oltarasenko:
I have addressed all the comments by now. Merging this, as I understand the logic itself looks appropriate.

PR #57, opened 2020-02-17
Add new fetch and parse functions

oltarasenko:
Change fetch function so it sends requests through all middlewares
Add parse function which allows to see how given page is going to be
parsed by a given spider.


oltarasenko:
I decided to force merge this code, as it's required to demonstrate another PR. I think that suggested changes are quite small & safe for the merge.

PR #59, opened 2020-02-17
Add browser rendering support

oltarasenko:
Added a fetcher with uses a Splash browser renderer in order to fetch
pages with it.

### DIFF
--- a/lib/crawly/fetchers/splash.ex
+++ b/lib/crawly/fetchers/splash.ex
@@ -1,0 +1,66 @@
+defmodule Crawly.Fetchers.Splash do
+  @moduledoc """
+  Implements Crawly.Fetchers.Fetcher behavior for Splash Javascript rendering.
+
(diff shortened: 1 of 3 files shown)


lib/crawly/fetchers/splash.ex
```diff
@@ -0,0 +1,66 @@
...
+          )
+
+      {base_url, other_options} ->
+        {base_url, other_options}
+      end
+
+    query_parameters = URI.encode_query(other_options)
+
+    url =
+      URI.merge(base_url, "?url=" <> request.url <> "&" <> query_parameters)
```
Ziinc: Should use https://hexdocs.pm/elixir/URI.html#encode_query/1 instead  for the url key. Maybe shift it to the query_parameters encoding expression

oltarasenko: +1, done

PR #60, opened 2020-02-25
Add Splash example

oltarasenko:
Add a link to an example project which shows how to extract data
from the site with dynamic content.


Ziinc:
My only concern with out-of-repo examples is that they get out of sync with the current version

oltarasenko:
I would prefer to keep the references to examples here. So people could try these projects. But I understand the concern. For now, I don't know a good way to solve it.

PR #61, opened 2020-02-26
Automatic cookies management

oltarasenko:


oltarasenko:
I will merge this now. It was online for quite a some time already. Also, it's a standalone middleware that can be always disabled. It should be safe to merge.

PR #71, opened 2020-03-27
Disable closespider_itemcount by default

oltarasenko:
This setting is mainly used for demo purposes. And it seems to be
highly confusing for a wide audience. We decided to disable it by
default, so crawly will not stop suddenly.


Ziinc:
Changes look good.

Issue #74, opened 2020-03-30
on_finish callback for spiders


AoRakurai:
I believe a on_finish/0 optional callback would be very beneficial (I personnaly need to know when my spiders finish and I would rather not be poking Crawly.Engine every X seconds to check if my spider is still running using Crawly.Engine.running_spiders).

It can be called if it is defined prior to 
    GenServer.call(__MODULE__, {:stop_spider, spider_name})
in Engine.

I have tested this change with my application and it seems to work, however I am very new to elixir and I don't know this repo well so I'm not certain this is an acceptable solution to this problem, and I'm not certain this would cover all cases of Spiders stoping.

Ziinc:
Seems easy for the Engine to run code that is beyond Crawly's context.

 What's wrong with polling the Engine for the Spider status? Seems like a good solution to me.

Another alternative is for the Engine to push messages on the spider status to a GenServer of choice.

oltarasenko:
@Ziinc I agree with @AoRakurai regarding the fact that it's better not to poll the status if it's possible. As I understand the desired function is to call a custom callback once the spider is exiting. It's similar to what we can find in [Scrapy](https://docs.scrapy.org/en/0.22/topics/signals.html). I was using it quite a lot when I was working with python and Scrapy, and I can agree it's useful. Maybe we should consider introducing all callbacks from the page above...

AoRakurai:
That would make Crawly a good fit for a lot more use cases I believe. The fact there was no on_finish callback forced me fork it temporarily so I could keep using it in my project.

Ziinc:
I agree having event callbacks are useful and that a concerted effort to cover most/all events would be good. However I don't agree with having this as a specific callback on the spider. 

If anything, this feels like something that should live in the settings, as an option. Then, with the custom_settings callback being implemented in #69 , hooking into the events would be as simple as doing
```
# assuming the eventually implemented callback expects a map of the settings... 
def custom_settings(global) do
    %{global | on_spider_stop: &MyApp.my_callback/1 } 
end
```
Where the my_callback function receives the spider state.

This allows for the dev to do a default callback, then override as necessary.

AoRakurai:
I agree that assuming settings can be implemented per spider, being able to define a on_spider_stop in settings would be ideal.

Ziinc:
Furthermore, I feel that these callback functions should be fired off using async Tasks, instead of synchronously. 

https://hexdocs.pm/elixir/Task.html

[...]

PR #78, opened 2020-04-07
Wrapper for Crawly.Spider behaviour

oltarasenko:
1. Extend Crawly.Spider with spider_closed callback
2. Use macro to simplify the Crawly.Spider definitions

### DIFF
--- a/lib/crawly/settings.ex
+++ b/lib/crawly/settings.ex
@@ -43,6 +43,7 @@ defmodule Crawly.Settings do
           # Defines retries
           retry: retry(),
           middlewares: [middleware()],
-          pipelines: [pipeline()]
+          pipelines: [pipeline()],
+          on_spider_closed_callback: function()
         ]
 end
--- a/lib/crawly/spider.ex
+++ b/lib/crawly/spider.ex
@@ -24,5 +24,13 @@ defmodule Crawly.Spider do
 
   @callback override_settings() :: Crawly.Settings.t()
 
-  @optional_callbacks override_settings: 0
+  defmacro __using__(_opts) do
+    quote do
+      @behaviour Crawly.Spider
+
+      def override_settings(), do: []
+
+      defoverridable override_settings: 0
+    end
+  end
 end
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ historical archival.
    ```elixir
    # lib/crawly_example/esl_spider.ex
    defmodule EslSpider do
-     @behaviour Crawly.Spider
+     use Crawly.Spider
+     
      alias Crawly.Utils
 
      @impl Crawly.Spider
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -105,6 +105,12 @@ default: 4001
 
 Allows to specify a custom port to start the application. That is important when running more than one application in a single machine, in which case shall not use the same port as the others.
 
+### on_spider_closed_callback :: function()
+
+default: :ignored
+
+Allows to define a callback function which will be executed when spider finishes
+it's work.
 
 ## Overriding global settings on spider level
 
--- a/documentation/introduction.md
+++ b/documentation/introduction.md
@@ -27,7 +27,7 @@ Here’s the code for a spider that scrapes blog posts from the Erlang Solutions
 
 ```elixir
 defmodule Esl do
-@behaviour Crawly.Spider
+  use Crawly.Spider
 
   @impl Crawly.Spider
   def base_url(), do: "https://www.erlang-solutions.com"
--- a/documentation/quickstart.md
+++ b/documentation/quickstart.md
@@ -24,7 +24,7 @@ Goals:
    ```elixir
    # lib/crawly_example/esl_spider.ex
    defmodule EslSpider do
-     @behaviour Crawly.Spider
+     use Crawly.Spider
      alias Crawly.Utils
 
      @impl Crawly.Spider
--- a/documentation/tutorial.md
+++ b/documentation/tutorial.md
@@ -75,7 +75,7 @@ homebase.ex under the lib/tutorial/spiders directory of your project.
 
 ```elixir
 defmodule Homebase do
-  @behaviour Crawly.Spider
+  use Crawly.Spider
 
   @impl Crawly.Spider
   def base_url(), do: "https://www.homebase.co.uk"
@@ -96,7 +96,7 @@ defmodule Homebase do
 end
 ```
 
-As you can see, our Spider implements the Spider behaviour and defines
+As you can see, our Spider implements the Crawly.Spider behaviour and defines
 some functions:
 
 1. base_url: method which returns base_urls for the given Spider, used in
@@ -267,7 +267,7 @@ into our spider.
 
 ```elixir
 defmodule Homebase do
-  @behaviour Crawly.Spider
+  use Crawly.Spider
 
   @impl Crawly.Spider
   def base_url(), do: "https://www.homebase.co.uk"
--- a/lib/crawly/engine.ex
+++ b/lib/crawly/engine.ex
@@ -21,9 +21,15 @@ defmodule Crawly.Engine do
     GenServer.call(__MODULE__, {:start_spider, spider_name})
   end
 
-  @spec stop_spider(module()) ::
-          :ok | {:error, :spider_not_running}
-  def stop_spider(spider_name) do
+  @spec stop_spider(module(), reason) :: result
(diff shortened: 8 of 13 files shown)


Ziinc:
I agree with the `use Crawly.Spider` macro, I think it would simplify boilerplate a lot.

I don't agree with the use of the `spider_closed/1` callback. This implies that there will be a callback for every spider-related event. I feel that this will give rise to a lot of boilerplate, and a lot of copy-pasting of code. There is also no access to the engine/spider state.

As I mentioned in the [issue](https://github.com/oltarasenko/crawly/issues/74#issuecomment-606708681), I lean towards simply passing a function for the engine to call  through the settings. It is very extensible and already allows for spider-level overrides.

Some benefits from this approach:
- The callback is fired asynchronously, which would be beneficial for other types of events
- can define a default callback, override on a per-spider basis as necessary
- can separate this callback from spider code (so cleaner spider code)
- less maintenance (less callbacks on the spider)
- easier to understand (less callbacks that may behave differently, this one is standardized to receive engine state and spider state)
- more extensible (the task firing can be abstracted to a util function, and adding support for other events would just be a few lines of code)

For example:
```elixir
# in config.exs
config :crawly,
  on_spider_close: &MyApp.send_email_notification/2

# in MyApp.MySpider
def override_settings() do
    [on_spider_close: &MyApp.handle_update_spider_dashboard/2 ]
end

# in Engine.close_spider
def stop_spider(spider_name, reason \\ :ignore) do
  callback = Crawly.Utils.get_settings(spider_name, :on_spider_close)

  # start the callback that we don't care about
   if callback do
     Supervisor.start_link([
        {Task, fn -> callback.(engine_state, spider_state) end}
      ], strategy: :one_for_one)

   end

  # close the spider
  GenServer.call(__MODULE__, {:stop_spider, spider_name})
end

```

oltarasenko:
>  don't agree with the use of the spider_closed/1 callback. This implies that there will be a callback for every spider-related event. I feel that this will give rise to a lot of boilerplate, and a lot of copy-pasting of code. There is also no access to the engine/spider state.
As I mentioned in the issue, I lean towards simply passing a function for the engine to call through the settings. It is very extensible and already allows for spider-level overrides.

I understand your concern (I was playing with both ways of declaring things before submitting the PR), and at this point, my arguments are the following.

**Custom settings way of defining callbacks have the following drawbacks**
 - custom_settings - is a KW list which allows overriding settings. **Semantically spider callbacks are not settings**. Even more, there is no global representation of the callbacks.
 - if we define for example spider_closed callback via settings, it would mean that our settings are more complex, now it's not only simple overrides of globally defined settings but also something on the top of it. 
- behavior itself will not be that clear: As well as callbacks are not defined by behavior, we would literally have to depend on conventions, which could be quite error-prone, when people will **do typos** and will not understand why something is not being called, and there will not be hints from the compiler. 
- finally, the Crawly internal code could be more simple. We would not have to do function_exported?/3 all the time. We can safely call something, expecting it to be defined default with the help of __using__ macro.

As far as I see the only benefit of custom_settings based approach is the fact that all spiders written for older versions of Crawly can be easily migrated without a switch to __using__ macro.

> Some benefits from this approach:

> The callback is fired asynchronously, which would be beneficial for other types of events
> can define a default callback, override on a per-spider basis as necessary

I think this can be easily achieved with the callback-based approach. We can force the macro to generate a function that will read the config by default.

> can separate this callback from spider code (so cleaner spider code)

Same as above, at this point more strict interface seems to be quite beneficial

> less maintenance (less callbacks on the spider)

Callbacks are not executed in the context of the spider. So technically it's just a place where the code is stored.

> more extensible (the task firing can be abstracted to a util function, and adding support for other events would just be a few lines of code)

We still can do it. E.g. the global setting might call a util function similarly to what you're suggesting.

Stating all the arguments above I would still prefer to go with the callback-based approach

Ziinc:
> I think this can be easily achieved with the callback-based approach. We can force the macro to generate a function that will read the config by default

But wouldn't this be going back one full circle to defining the callback through settings? 

I agree that the callback isn't settings. But neither is it spider code. It shouldn't be living in the spider as well. These type of event-triggered callbacks are side-effects and are not related to data-parsing, which is the context of a spider module. 

From a user story perspective:
1. As a dev, I want to do something when this spider ends -- spider specific
2. As a dev, I want to do something when any of my spider ends -- not spider specific
3. As a dev, I want to do something when all my spiders end. -- not spider specific

The settings-approach allows for addressing of scenarios 1,2,3. While the macro callback approach only addresses scenario 1.

Relying on the macro to define the callbacks in advance also makes it a breaking change, and requires internal code to check if the callback is defined anyway.

Why have 2 different approaches to configuring a Spider when there already is a unified approach to configure it? In the end, if there's going to be a default setting on the config, it can be customized with `override_settings` and makes the callback moot.

Ziinc:
I feel that the KW approach forces the dev to define the callback outside of the spider scope, which would be more appropriate for scenarios 2 and 3.

Ziinc:
I also think the typo and compiler concerns can be rectified through debug logging.

I think the crux of the issue here is:
1. where should the callback code be defined, in spider or elsewhere
    - if in spider, then macro callback is an ok approach
    - if not in spider, then define elsewhere and point to it in settings.
2. How is this type of configuration (event callbacks) going to be extended in the future.
   - macro approach means 1 callback per event
   - KW setting approach means one new setting per event
3. How to specify a default (i.e. global) callback?

oltarasenko:
> But wouldn't this be going back one full circle to defining the callback through settings?
Maybe not, if the settings are clearly defined and the callbacks code is autogenerated with a macro

> I agree that the callback isn't settings. But neither is it spider code. It shouldn't be living in the 
 spider as well. These type of event-triggered callbacks are side-effects and are not related to data- parsing, which is the context of a spider module.

Mmm.. Yes, you're right. Maybe they should not be a part of the spider at all. Let me ask for an advise from Scrapy core team members.

> Relying on the macro to define the callbacks in advance also makes it a breaking change,
Yes, it's a breaking change. 

> and requires internal code to check if the callback is defined anyway.
well, this is going to be handled by the compiler, who will return a warning stating that you have not defined required callbacks.

> Why have 2 different approaches to configuring a Spider when there already is a unified approach to configure it? In the end, if there's going to be a default setting on the config, it can be customized with override_settings and makes the callback moot.

Well, but from the other side, I see some similarity between parse_item/1 and spider_closed/1 or item_scraped/2 [item scraped would be even more specific to a given spider, e.g. it would allow to make some very specific post-processings]

oltarasenko:
@Ziinc Actually after some conversations, I have decided to switch to the idea of the settings_based approach.

Please have another glance on the PR.

lib/crawly/spider.ex
```diff
@@ -13,6 +13,9 @@ defmodule Crawly.Spider do
       provide custom spider specific settings. Should define a list with custom
       settings and their values. These values will take precedence over the
       global settings defined in the config.
+  5. `spider_closed/1` an optional callback which can be used in order to add
```
Ziinc, line 16: Not used anymore, can remove

lib/crawly/spider.ex
```diff
@@ -23,6 +26,17 @@ defmodule Crawly.Spider do
               Crawly.ParsedItem.t()
 
   @callback override_settings() :: Crawly.Settings.t()
+  @callback spider_closed(reason) :: :ok
```
Ziinc, line 29: Not used

lib/crawly/spider.ex
```diff
@@ -23,6 +26,17 @@ defmodule Crawly.Spider do
...
+            when reason: atom()
 
-  @optional_callbacks override_settings: 0
+  defmacro __using__(_opts) do
+    quote do
+      @behaviour Crawly.Spider
+
+      def override_settings(), do: []
+      def spider_closed(_reason), do: :ok
```
Ziinc, line 37: Not used

lib/crawly/settings.ex
```diff
@@ -43,6 +43,7 @@ defmodule Crawly.Settings do
           # Defines retries
           retry: retry(),
           middlewares: [middleware()],
-          pipelines: [pipeline()]
+          pipelines: [pipeline()],
+          on_spider_closed_callback: function()
```
Ziinc, line 47: Do you think it'll be good to nest it under an `events` or `callbacks` key? For organisational purpose

oltarasenko: I am not quite sure, from one side I don't want to have nested sections, as the configuration will be more complex. Lets maybe keep it as it is for now, and see if we need to tweak later.

PR #85, opened 2020-04-17
Update tutorial.md

michaltrzcinka:
Hello,

I went through the tutorial and updated it. 
There's one issue though - I can't see any "Scraped: ..." messages when running the spider. Has that been removed on purpose or is it a bug?

### DIFF
--- a/documentation/tutorial.md
+++ b/documentation/tutorial.md
@@ -178,47 +178,7 @@ You will see something like:
     {"Pragma", "no-cache"},
     {"Expires", "-1"},
     {"Vary", "Accept-Encoding"},
-    {"Set-Cookie", "Bunnings.Device=default; path=/"},
-    {"Set-Cookie",
-     "ASP.NET_SessionId=knvfhxsyghuqtrki0jgn4put; path=/; HttpOnly; SameSite=Lax"},
-    {"Set-Cookie", "Bunnings.Device=default; path=/"},
-    {"Set-Cookie",
-     "ASP.NET_SessionId=knvfhxsyghuqtrki0jgn4put; path=/; HttpOnly; SameSite=Lax"},
-    {"Set-Cookie", "Bunnings.UserType=RetailUser; path=/"},
-    {"Set-Cookie",
-     "Bunnings.NearestStoreID=335; expires=Sat, 17-Apr-2021 10:34:35 GMT; path=/"},
-    {"Set-Cookie",
-     "Bunnings.NearestStoreName=Milton Keynes; expires=Sat, 17-Apr-2021 10:34:35 GMT; path=/"},
-    {"Set-Cookie",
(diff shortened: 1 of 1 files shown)


Ziinc:
I think the logging was updated during the previous version's refactoring.

Could you give a shot to automate the writing of the tutorial as part of a test module, so that it's always in sync?

Ziinc:
You can try using eex templates

documentation/tutorial.md
```diff
@@ -36,8 +36,6 @@ This will create a tutorial directory with the following contents:
 ```bash
 tutorial
 ├── README.md
-├── config
```
oltarasenko, line 39: Maybe this should not be removed?

documentation/tutorial.md
```diff
@@ -159,54 +157,85 @@ please hold on. We're going to cover it in the next section.
...
-   status_code: 200
- }}
+%HTTPoison.Response{
+  body: "[response body here...]"
+  headers: [
+    {"Date", "Fri, 17 Apr 2020 10:34:35 GMT"},
+    {"Content-Type", "text/html; charset=utf-8"},
+    {"Transfer-Encoding", "chunked"},
+    {"Connection", "keep-alive"},
+    {"Set-Cookie",
```
oltarasenko, line 175: I am not sure we need all these Set-Cookie everywhere. Please remove some of them, so the tutorial is shorter

documentation/tutorial.md
```diff
@@ -332,7 +361,24 @@ end
...
+    {Crawly.Pipelines.WriteToFile, folder: "/tmp", extension: "csv"}
+  ]
+```
+The scraped data will now be stored in a CSV file under `/tmp` directory on your filesystem. The name of the file
+will be the same as our spider name - in our case `Homebase.csv`.
+
+If you restart iex and run this spider `Crawly.Engine.start_spider(Homebase)` again, it will output the extracted data 
+with the log:
+
+TODO: I can't see any "Scraped" messages like these below. Has that been removed on purpose or is a bug?
```
oltarasenko, line 381: I think you're right, please remove these messages, as they do not show up anymore

oltarasenko:
@Ziinc I think that current PR improves the document. We can accept it. Regarding the automation, let's discuss it in a separate commit/ticket. Could you maybe explain how it might be done as I am a bit puzzled?
@michaltrzcinka Good job. Some minor fixes please!

documentation/tutorial.md
michaltrzcinka: `mix new` doesn't create this directory anymore, that's why I removed it here and added instructions to create the config file manually.

documentation/tutorial.md
michaltrzcinka: Sure, I will do that.

documentation/tutorial.md
michaltrzcinka: 👍 will do

PR #88, opened 2020-04-23
DownloadMedia pipeline

michaltrzcinka:
Hi, this is the first naive attempt to create a pipeline that will download resources. There are no tests or documentation yet - just wanted to check if this is a good direction.

A couple of questions:

- This pipeline doesn't transform items - maybe it should be a middleware?
- Is it OK to download directly in the pipeline? Shouldn't we do it asynchronously in other processes?
- Is it OK to keep full response body in memory and only then pass it to `File.write`? Maybe we could do some sort of streaming?

### DIFF
--- a/lib/crawly/asset_downloader.ex
+++ b/lib/crawly/asset_downloader.ex
@@ -1,0 +1,44 @@
+defmodule Crawly.AssetDownloader do
+  require Logger
+
+  use GenServer
+
+  def start_link([]) do
+    Logger.debug("Starting asset downloader")
+
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def init(_args) do
+    {:ok, %{}}
+  end
+
+  def download_asset(directory, url) do
+    GenServer.cast(__MODULE__, {:download_asset, directory, url})
+  end
+
+  @impl true
+  def handle_cast({:download_asset, directory, url}, state) do
+    :ok = File.mkdir_p(directory)
+
+    case HTTPoison.get(url) do
+      {:ok, response} ->
+        full_path = Path.join(directory, file_name(url))
+        :ok = File.write(full_path, response.body)
+        Logger.debug("Downloaded image from URL #{url} into #{full_path}")
+      {:error, error} ->
+        Logger.error("Could not download image from URL #{url}: #{inspect(error)}")
+    end
+
+    {:noreply, state}
+  end
+
+  # Returns the string after the last slash, e.g.:
+  # https://amazon.com/products/picture.jpg => picture.jpg
+  # It is usually a good approximation of a "file name" on the web
+  defp file_name(url) do
+    Regex.run(~r/(?<=\/)[^\/]+$/, url)
+    |> List.first()
+  end
+end
--- a/lib/crawly/application.ex
+++ b/lib/crawly/application.ex
@@ -9,11 +9,12 @@ defmodule Crawly.Application do
     import Supervisor.Spec, warn: false
     # List all child processes to be supervised
 
-    children = [
+      children = [
       worker(Crawly.Engine, []),
       supervisor(Crawly.EngineSup, []),
       {Crawly.DataStorage, []},
       {Crawly.RequestsStorage, []},
+      {Crawly.AssetDownloader, []},
       {DynamicSupervisor,
        strategy: :one_for_one,
        name: Crawly.RequestsStorage.WorkersSup},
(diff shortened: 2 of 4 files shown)


michaltrzcinka:
@Ziinc @oltarasenko not sure if you're getting notifications about draft pull requests ;) please take a look

oltarasenko:
Hi @michaltrzcinka. Sorry I did not have a chance to look at your PR earlier. 

I don't like the idea of putting the image downloading code into the item pipeline. The pipeline is executed in the context of the DataStorageWorker which is one process (bottleneck in this case). The overall idea was to use the Item pipelines for relatively lightweight data transformations and image downloading is a problem in this case.

I would try to do it with a context of a worker instead (read in `parse_item`). Otherwise, another approach would be to run a separate process, so the pipeline is just sending urls to it.

Ziinc:
Perhaps an additional `init` callback on the pipeline behaviour would allow the necessary setup to be done, while containing all file downloading code in one place. 

The worker could iterate over the pipelines at the beginning of the scrape, letting the file downloader pipeline setup its own queue and store process references in the state.

When the worker pipes the scraped item, the file downloader pipeline can then just forward the url to the downloading process.

oltarasenko:
> Perhaps an additional init callback on the pipeline behaviour would allow the necessary setup to be done, while containing all file downloading code in one place.
> 
> The worker could iterate over the pipelines at the beginning of the scrape, letting the file downloader pipeline setup its own queue and store process references in the state.

It would prefer to avoid starting a process like that. Especially taking into account that it will not be supervised. We can have a special AssetsDownloader process (or pool of processes) started on Crawly's startup. In this case, the given image downloader pipeline can just get a reference to the name of this process.

michaltrzcinka:
I introduced `AssetDownloader`, WDYT?

oltarasenko:
@michaltrzcinka It can be used as a sort of starting point. E.g. we can fine-tune it in some way. But could you do a more complete example, so I can see it storing images?

Another request. Could you try downloading images from `parse_item/1` callback

lib/crawly/asset_downloader.ex
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    {:noreply, state}
+  end
+
+  # Returns the string after the last slash, e.g.:
+  # https://amazon.com/products/picture.jpg => picture.jpg
+  # It is usually a good approximation of a "file name" on the web
+  defp file_name(url) do
+    Regex.run(~r/(?<=\/)[^\/]+$/, url)
```
oltarasenko, line 41: I would suggest storing images using guid as a file name. Otherwise we will get duplicates and overrides. It's better to extend item with a field called downloaded images
.

michaltrzcinka:
> @michaltrzcinka It can be used as a sort of starting point. E.g. we can fine-tune it in some way. But could you do a more complete example, so I can see it storing images?
> 
> Another request. Could you try downloading images from `parse_item/1` callback, to have it as an example before we will finish with Accets Downloader

When you run [this example](https://github.com/oltarasenko/crawly/pull/88/files#diff-0c038908c088c01420d1e2ecab8e0b85), it will actually run through the new `DownloadMedia` pipeline which in turn calls `AssetDownloader`. Or do you mean a different kind of example?

lib/crawly/asset_downloader.ex
michaltrzcinka: I don't understand the last sentence, could you please explain?

oltarasenko: I suggest generating image filenames in the pipeline(`download_media.ex`) so the assets downloader will get two things
1. URL to download from
2. path to store

The item pipeline could look this way in the case:
```
def run(item, state, opts \\ []) do
.....
AssetDownloader.download_asset(filename, image_url)
{Map.put(item, :image_path, filename) , state}
end
```
In this case, the item structure will already have the field image_path which will point to the actual image.

oltarasenko:
Yeah, sorry I wanted to have an example without AssetsDownloader code. As a separate work.

Ziinc:
I've been thinking about this. Would it be better to have a generalised SupervisedQueuedWork (just a tentative name)  process that developers can hook into?

This allows us to queue any type of long running functions for the process to carry out.

Besides asset downloading, I can think of other use cases, like file uploading, data validation, streaming.

The function to do the asset downloading can be defined in the pipeline, then the pipeline, when runned, can hand off the the function identifier and arguments to the queue to be executed in a FIFO fashion.

This makes the supervised process more reusable, instead of it being solely for asset downloading.

PR #97, opened 2020-05-08
Crawly UI experimental integration

oltarasenko:
This PR contains changes required to make integration of Crawly & CrawlyUI possible.

The code is currently considered experimental. I am testing it quite intensively on related demo instances. However, I would take it into the codebase as a part of the experimental namespace.


Ziinc:
Having it as a separate node just for serving the UI seems like a lot of additional overhead. What are the benefits of doing so?

Could we achieve the same end goal (viewing manager stats in real time) through a Cowboy/phoenix process on the engine node? All stats (crawl speeds, request storage count, parsed items count) are already on the manager process as well.

I also don't think that the http endpoint would be so resource heavy as to require a separate node. After all, there is already the basic http api endpoint

Ziinc:
Additionally, it would make more sense to break up the UI into modular pieces, to facilitate integration into existing code bases.

It would be more likely that the end user of the ui already had an admin dashboard. Instead of having separate dedicated endpoints, perhaps the ui could be in html fragments/iframes that can be embedded into an existing dashboard. For example, if I were to integrate the ui into my own codebase, do a partial integration of the ui, as I already have an existing dashboard (with auth etc).

Ziinc:
Perhaps a conditional Boolean setting for enabling the ui would be enough? Where it would conditionally start up a process from the crawly_ui package

oltarasenko:
> Having it as a separate node just for serving the UI seems like a lot of additional overhead. What are the benefits of doing so?
> Could we achieve the same end goal (viewing manager stats in real time) through a Cowboy/phoenix process on the engine node? All stats (crawl speeds, request storage count, parsed items count) are already on the manager process as well.
> Additionally, it would make more sense to break up the UI into modular pieces, to facilitate integration into existing code bases.

> It would be more likely that the end user of the ui already had an admin dashboard. Instead of having separate dedicated endpoints, perhaps the ui could be in html fragments/iframes that can be embedded into an existing dashboard. For example, if I were to integrate the ui into my own codebase, do a partial integration of the ui, as I already have an existing dashboard (with auth etc).

Definitely there is an overhead, however, let me explain my motivation here:
1. The approach above gives the possibility to run multiple crawly instances on separate nodes from one UI. This is done in order to separate load on multiple instances, otherwise, bandwidth is a bottleneck, at least for us :(.
2.  I am not a great fan of the idea of adding Phoenix to Crawly repo. It looks like overkill, from the dependency's point of view. I want crawly to be simple and extendible. I know other people are building GRPC based frontend for Crawly. I simply don't want to tell how to build it! As soon as Crawly can just route items somewhere anyone (anything) can be Crawly's frontend.
3. Scaling - yeah I am not even sure that the current postgres based approach will work for our case. E.g. we have millions of items to be inserted per job :(, maybe, in this case, I will switch towards Mongo. Again when I am thinking about having postgres (and potentially Mongo) based deps in Crawly's codebase, I feel like it's not the best place for them to be.

To summarize: 
1. I kind of want to intersect my current production requirements (being able to run massively distributed crawls)  with Crawly/CrawlyUI, so I can have a bit more time to spend on the project. 
2. Some users have started to build their own private GRPC based UI. And I support the idea that it should be possible to build more than one UI for the tool
3. I don't want to bring lots of dependencies to Crawly. E.g. I don't want Crawly to create/manage databases, etc


Hopefully, my arguments were convincing enough!

Ziinc:
I think the crawly_ui project is not actually concerned about just the ui. It implements both data storage and the ui layer. In my opinion, it should be split up into a crawly_repo (database & migrations only) and crawly_ui (live crawling, can hook into the repo). 
This makes it more useful for users who are doing flat-file crawling and just want to see live stats. It also makes the storage part optional, for people who don't want to manage their own distributed storage solution. 


I do admit that the multi-node option is good for distributed crawling, but it won't be needed for many users who are not at that scale. Setting up another node shouldn't be a requirement in order to use the ui solution. 

I was thinking along the lines of crawly_ui hooking into crawly engine and only implementing the phoenix areas, live streaming of manager/parsed items. This keeps phoenix out of the crawly repository.

Ziinc:
I have heard of coackroachdb as an postgres alternative for geo distribution. However I don't see why postgres can't handle that amount, all that I'd required is adding more slave nodes.

I'd advise against nosql databases,unless you don't need consistency/normalisation. And even then I'd pick mnesia over mongo.

Ziinc:
Also, I'm not saying that we should add the database and ui as dependencies in the crawly repo. They can be installed separately and inter operate through macros.

oltarasenko:
@Ziinc I've got your point! Actually now I like the idea of also adding a lightweight web just to manage crawly (start/stop/speed/list), without the storage.

I am still a bit concerned about adding something as big as phoenix here, as Phoenix is UI + database in most of the cases. However, we can think & brainstorm it. My colleagues have shown me very very nice command-line based UX:  https://github.com/ndreynolds/ratatouille#building-an-application. 

Let's agree on the following - let's have a separate task for lightweight CrawlyCommander for people who do not need UI, but let's also merge this experimental middleware, so I can unblock my work on the project using Crawly.

lib/crawly/utils.ex
```diff
@@ -139,6 +139,31 @@ defmodule Crawly.Utils do
...
+  @spec list_spiders() :: [module()]
+  def list_spiders() do
+    Enum.reduce(
+      get_modules_from_applications(),
+      [],
+      fn mod, acc ->
+        try do
+          beh = Keyword.get(mod.__info__(:attributes), :behaviour)
+
+          case beh == [Crawly.Spider] do
```
Ziinc, line 155: should use Enum.any?/2 check instead, in case a module implements multiple (custom) behaviours

lib/crawly/utils.ex
```diff
@@ -139,6 +139,31 @@ defmodule Crawly.Utils do
...
+  @doc """
+  Returns a list of known modules which implements Crawly.Spider behaviour
+  """
+  @spec list_spiders() :: [module()]
+  def list_spiders() do
+    Enum.reduce(
+      get_modules_from_applications(),
+      [],
+      fn mod, acc ->
+        try do
```
Ziinc, line 152: Why is the try/rescue needed here? Its an anti-pattern for swallowing errors.

lib/crawly/pipelines/experimental/send_to_ui.ex
```diff
@@ -0,0 +1,42 @@
...
+          UUID.uuid1()
+
+        tag ->
+          tag
+      end
+
+    ui_node =
+      case Keyword.get(opts, :ui_node) do
+        nil ->
+          throw(
```
Ziinc, line 22: Instead of throwing, it would be better to gracefully do nothing and log errors instead.

Throwing would cause the entire pipe to fail

lib/crawly/pipelines/experimental/send_to_ui.ex
```diff
@@ -0,0 +1,42 @@
...
+  @moduledoc """
+  """
+  @behaviour Crawly.Pipeline
+
+  require Logger
+
+  @impl Crawly.Pipeline
+  def run(item, state, opts \\ []) do
+    job_tag =
+      case Map.get(state, :job_tag, nil) do
```
Ziinc, line 11: I'd refactor this to:
```
state
|> Map.get(:job_tag, UUID.uuid1())
...
```

Ziinc: I'd also use a with/do/else instead of a having multiple case/do when all you're doing is binding variables.

documentation/experimental_ui.md
```diff
@@ -0,0 +1,78 @@
...
+
+In order to try it:
+1. clone crawly_ui repo: `git clone git@github.com:oltarasenko/crawly_ui.git`
+2. build ui and worker nodes: `docker-compose build`
+3. apply migrations: `docker-compose run ui bash -c "/crawlyui/bin/ec eval \"CrawlyUI.ReleaseTasks.migrate\""`
+4. run it all: `docker-compose up`
+
+## Live demo
+
+Live demo is available as well. However it might be a bit unstable due to continues
```
Ziinc, line 57: our continuous release process.

documentation/experimental_ui.md
```diff
@@ -0,0 +1,78 @@
...
+releases process. Please give it a try and let us know what do you think
+
+[Live Demo](http://18.216.221.122/)  
+
+## Items browser
+
+One of the cool features of the CrawlyUI is items browser which allows comparing
+extracted data with a target website loaded in the IFRAME. However due to the
+fact that most of the big sites would block iframes, it will not work for you.
+Unless you install special browser extension to ignore X-Frame headers. For example
```
Ziinc, line 67: However, as sites may block iframes, a workaround browser extension may be used to ignore X-Frame headers.

documentation/experimental_ui.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. Schedule (start and stop) your spiders on a cloud
+2. View running jobs (performance based analysis)
+3. View and validate scraped items for quality assurance and data analysis purposes.
+4. View individual items and compare them with the actual website.
+
+## Project status
+
+Currently the project is in early alpha stage. We're constantly working on making
+it more stable. And currently running it for long running jobs.
```
Ziinc, line 18: ~~And currently running it for long running jobs.~~

Ziinc: ~~so far it did no show major problems, however we accept that there will be problems on such early stages! If you are one of those who has them, please don't 
hesitate to report them here.~~

should report it on the crawly_ui repo, not crawly.

README.md
```diff
@@ -97,6 +97,20 @@ of asynchronous elements (for example parts loaded by AJAX).
 You can read more here:
 - [Browser Rendering](https://hexdocs.pm/crawly/basic_concepts.html#browser-rendering)
 
+## New: Experimental UI
```
Ziinc, line 100: ~~New:~~ Experimental UI

README.md
```diff
@@ -97,6 +97,20 @@ of asynchronous elements (for example parts loaded by AJAX).
...
+
+We have made few first steps towards building UI for Crawly project which would
```
Ziinc, line 102: should be more concise.

The CrawlyUI project is an add-on that aims to provide an interface for managing and rapidly developing spiders.

Ziinc: can replace the whole paragraph with this line. They say the same things

documentation/experimental_ui.md
```diff
@@ -0,0 +1,78 @@
+# Experimental UI
+---
+
+We believe that web scraping is a process. It might seem easy to extract first
```
Ziinc, line 4: The [CrawlyUI project] (link to repo) is an additional opt-in user interface for managing and developing Crawly spiders.

Don't need to repeat the philosophy of crawly

Ziinc: docs should be short and to the point, cos people will be skimming through them

Ziinc:
> Let's agree on the following - let's have a separate task for lightweight CrawlyCommander for people who do not need UI, but let's also merge this experimental middleware, so I can unblock my work on the project using Crawly.

Yes, i think Ratatouille looks very nice, though i think it can be an opt-in tui layer as well for people doing flatfile crawling, just like how the web ui would be opt in. Exposing Elixir-level apis at the Engine/commander(?) level would suffice for this repo. Also, not too sure about the name commander. Hmmm maybe Conductor? Orchestrator? Chief? 

While we're at it, we should also probably standardize the http api endpoint to adhere to the same management capabilities.

Ziinc:
Oh yea, to  make development and testing easier, I think including the any future ui/db repos as submodules inside a `extras` folder might be a good idea, to keep everything in sync

lib/crawly/utils.ex
oltarasenko: Nice catch. I have added a fix & test for it.

lib/crawly/utils.ex
oltarasenko: Yes, it is. Honestly, I don't like the current approach for getting spiders. But also, I don't like to introduce something like SPIDERS_DIR option to crawly config. At least not yet. 

I took the approach above from a snippet on the internet. It crashes in some cases for example:
```
09:09:35.756 [debug] Could not get behaviour information for: %UndefinedFunctionError{arity: 1, function: :__info__, message: nil, module: :cowboy_websocket, reason: nil}

09:09:35.832 [debug] Could not get behaviour information for: %UndefinedFunctionError{arity: 1, function: :__info__, message: nil, module: :plug_multipart, reason: nil}
```
When it comes to erlangish deps. However, it works for Crawly.Spider behaviour. Probably in future, we need a better way of finding spiders.

lib/crawly/pipelines/experimental/send_to_ui.ex
oltarasenko: 👍

lib/crawly/pipelines/experimental/send_to_ui.ex
oltarasenko: good!

documentation/experimental_ui.md
oltarasenko: 👍

documentation/experimental_ui.md
oltarasenko: 🆗

documentation/experimental_ui.md
oltarasenko: ok, agree

README.md
oltarasenko: Thanks!

oltarasenko:
@Ziinc thanks for looking into it. I think I have addressed all your comments now.

lib/crawly/utils.ex
Ziinc: I've got a snippet that I use for finding spider modules that is regex based. I'll dig it out from my code base to see if it helps

oltarasenko: thanks, but maybe lets do it as a separate PR. This approach currently works.

Issue #91, opened 2020-04-27
WriteFile folder option improvements


Vesuvium:
Hi everyone! I've been using crawly recently and I found the folder option a bit confusing. 

The folder is always set as `/tmp` which makes think that only absolute paths are allowed. A single example with a local or `~` path would make things clearer, for example here in the last one: https://hexdocs.pm/crawly/Crawly.Pipelines.WriteToFile.html

The other thing that bugs me is that the folder has to exist. It would be nicer if the folder is created when missing. This would open the possibility of setting the default to a local path, making it more immediate whether the parser is working

Ziinc:
It actually defaults to the current folder.

Documentation should be fixed

Ziinc:
And the folder option is optional, it doesn't always need to be declared.

oltarasenko:
Released

PR #98, opened 2020-05-12
Improve folder creation in WriteToFile item pipeline

oltarasenko:
As it's requested in https://github.com/oltarasenko/crawly/issues/91

Now the pipeline will also create a folder on the filesystem
if it does not exist.

Also improved documentation.


test/pipelines/write_to_file_test.exs
```diff
@@ -77,4 +79,83 @@ defmodule Pipelines.WriteToFileTest do
...
+    )
+
+    :meck.expect(
+      NaiveDateTime,
+      :to_string,
+      fn _ -> ts end
+    )
+
+    pipelines = [
+      {Crawly.Pipelines.WriteToFile, folder: "/tmp", extension: "csv", include_timestamp: true}
```
Ziinc, line 144: if it is by default true, should test for when the flag is false

test/pipelines/write_to_file_test.exs
```diff
@@ -77,4 +79,83 @@ defmodule Pipelines.WriteToFileTest do
...
+    item = @binary
+
+    state = %{spider_name: MySpider}
+
+    # run the pipeline
+    _result = Crawly.Utils.pipe(pipelines, item, state)
+
+    receive do
+      msg ->
+        assert String.contains?(msg, ts)
```
Ziinc, line 156: shift the timestamp assertion up to the csv test, as it is a default behaviour

lib/crawly/pipelines/write_to_file.ex
```diff
@@ -57,22 +62,27 @@ defmodule Crawly.Pipelines.WriteToFile do
...
-    extension =
-      Map.get(opts, :extension, "jl")
+    extension = Map.get(opts, :extension, "jl")
 
-    fd = open_fd(state.spider_name, folder, extension)
+    filename = case  Map.get(opts, :include_timestamp, false) do
+      false -> "#{inspect(state.spider_name)}.#{extension}"
+      true ->
+        ts_string = NaiveDateTime.utc_now() |> NaiveDateTime.to_string()
+        "#{inspect(state.spider_name)}_#{ts_string}.#{extension}"
```
Ziinc, line 78: i suggest using a `DateTime.to_unix/1` instead, as there are no special characters

oltarasenko: Are you talking about something like this:
```
 datetime = DateTime.now!("Etc/UTC") |> DateTime.to_unix |> to_string
"1589443961"
```
In general, we can do it, but it will be a bit less clear for the end-user?

test/pipelines/write_to_file_test.exs
oltarasenko: done.

test/pipelines/write_to_file_test.exs
oltarasenko: Done. I have moved the test higher and added a test to verify that TS is not added if it's not expected

oltarasenko:
I have addressed all comments except the one related `to_unix`. Please explain it a bit more, so I can adjust my code as well.

lib/crawly/pipelines/write_to_file.ex
Ziinc: I agree. I think perhaps a nice solution would be to replace characters such that characters like #, space, :,. with underscores so that they don't make it difficult for doing scripting

oltarasenko: Sorry could you explain what should I do it terms of the code, as I am a bit confused.

Ziinc:
```
 ts_string = NaiveDateTime.utc_now() 
|> NaiveDateTime.to_string()
|> String.replace(" .,#:", "_")
```

oltarasenko: hm... I was not aware NaiveDateTime.utc_now()  could produce it..

Ziinc: Oh sorry just checked the test assertion, saw wrongly.

```
# in the test
 ts = "2020-05-13 09:06:22.668828"
```
so it should replace the hyphens, colons, full stops, and spaces instead.
string replacement pattern should be " -.,:"

oltarasenko: Please check

Ziinc: Looks good

PR #104, opened 2020-05-18
added Crawly.Engine.list_known_spiders/0

Ziinc:
Adds a `list_known_spiders/0`, `refresh_spider_list/0` functions to `Crawly.Engine`. 

Expected behaviour: 
- lists all started/stopped spiders and their pid if available.

Behaviour intersects with `Crawly.Engine.running_spiders/0`, it may be a good idea to remove `running_spiders/0` and add in a filter option for `list_spiders/0`

### DIFF
--- a/lib/crawly/engine.ex
+++ b/lib/crawly/engine.ex
@@ -10,6 +10,9 @@ defmodule Crawly.Engine do
 
   @type t :: %__MODULE__{started_spiders: started_spiders()}
   @type started_spiders() :: %{optional(module()) => identifier()}
+  @type list_spiders() :: [
+          %{name: module(), state: :stopped | :started, pid: identifier()}
+        ]
 
   defstruct started_spiders: %{}
 
@@ -33,6 +36,11 @@ defmodule Crawly.Engine do
     GenServer.call(__MODULE__, {:stop_spider, spider_name})
   end
 
+  @spec list_spiders() :: list_spiders()
+  def list_spiders() do
+    GenServer.call(__MODULE__, :list_spiders)
+  end
+
   @spec running_spiders() :: started_spiders()
   def running_spiders() do
     GenServer.call(__MODULE__, :running_spiders)
@@ -49,6 +57,24 @@ defmodule Crawly.Engine do
 
   def handle_call(:running_spiders, _from, state) do
     {:reply, state.started_spiders, state}
+  end
+
+  def handle_call(:list_spiders, _from, state) do
+    list =
+      Crawly.Utils.list_spiders()
+      |> Enum.map(fn name ->
+        %{
+          name: name,
+          state:
+            case Map.has_key?(state.started_spiders, name) do
+              true -> :started
+              false -> :stopped
+            end,
+          pid: Map.get(state.started_spiders, name)
+        }
+      end)
+
+    {:reply, list, state}
   end
 
   def handle_call({:start_spider, spider_name}, _form, state) do
--- a/lib/crawly/engine.ex
+++ b/lib/crawly/engine.ex
@@ -8,13 +8,19 @@ defmodule Crawly.Engine do
 
   use GenServer
 
-  @type t :: %__MODULE__{started_spiders: started_spiders()}
+  @type t :: %__MODULE__{
+          started_spiders: started_spiders(),
+          known_spiders: [module()]
+        }
   @type started_spiders() :: %{optional(module()) => identifier()}
-  @type list_spiders() :: [
-          %{name: module(), state: :stopped | :started, pid: identifier()}
-        ]
 
-  defstruct started_spiders: %{}
+  @type spider_info() :: %{
+          name: module(),
+          status: :stopped | :started,
+          pid: identifier() | nil
+        }
+
+  defstruct(started_spiders: %{}, known_spiders: [])
 
   @spec start_spider(module()) ::
           :ok
@@ -26,7 +32,8 @@ defmodule Crawly.Engine do
 
   @spec stop_spider(module(), reason) :: result
         when reason: :itemcount_limit | :itemcount_timeout | atom(),
-             result: :ok | {:error, :spider_not_running}
+             result:
+               :ok | {:error, :spider_not_running} | {:error, :spider_not_found}
   def stop_spider(spider_name, reason \\ :ignore) do
     case Crawly.Utils.get_settings(:on_spider_closed_callback, spider_name) do
       nil -> :ignore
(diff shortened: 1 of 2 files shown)


lib/crawly/engine.ex
```diff
@@ -87,4 +99,19 @@ defmodule Crawly.Engine do
 
     {:reply, msg, %Crawly.Engine{state | started_spiders: new_started_spiders}}
   end
+
+  defp list_all_spider_status(started_spiders) do
+    Crawly.Utils.list_spiders()
```
oltarasenko, line 104: Please move this call to list_spiders() on line 40. As I see it we don't need to do it within the context of the Engine process.

lib/crawly/engine.ex
```diff
@@ -10,6 +10,9 @@ defmodule Crawly.Engine do
 
   @type t :: %__MODULE__{started_spiders: started_spiders()}
   @type started_spiders() :: %{optional(module()) => identifier()}
+  @type list_spiders() :: [
```
oltarasenko, line 13: I would suggest to do it slightly differently:

```
@type spider_status() :: :stopped | {:started, identifier()}

# And in the function we could have
@spec list_spiders() :: [%{name: module(), state: spider_status()}]
```

I prefer this type of specs...

test/engine_test.exs
```diff
@@ -0,0 +1,21 @@
+defmodule EngineTest do
+  use ExUnit.Case
+
+  test "list_spiders/0 lists all spiders and their current status in the engine" do
+    assert spiders = Crawly.Engine.list_spiders()
+    assert [_ | _] = spiders
```
oltarasenko, line 6: What happens here? Just in case I am a bit worried about the case when we add more test spiders... Will this test =fail automatically in this case?

lib/crawly/engine.ex
Ziinc: OK will adjust

lib/crawly/engine.ex
Ziinc: I moved it to a priv function as the stop_all_spiders and (I think) start_all_spiders uses it to check if a spider should be started/stopped

test/engine_test.exs
Ziinc: I'll remove the unnecessary assertion on line 5, it should never fail.

lib/crawly/engine.ex
oltarasenko: I was asking to move `Crawly.Utils.list_spiders()` out from the `handle_call` scope. It's potentially a heavy operation (and I saw erlang nodes killed by requests to list all applications). So engine will not be blocked by the listing process.

Another idea which sits in my mind is having a registry of spiders in the Engine. So when engine starts we can do `Crawly.Utils.list_spiders()` so it will know which spiders it have, and will store this info in the state. With the idea that if you want to update the list you can do for example `Engine.scan_spiders/0` or `Engine.add_spider`

oltarasenko:
@Ziinc should I sketch the idea of having the list of spiders in the state? I have a big chunk of time now, so I can do it fast.

Ziinc:
using https://hexdocs.pm/elixir/Registry.html ?
It is feasible to keep the list of spiders in state, i don't think we will need to refresh it unless devs do hot code swapping. Should only need to call it once in `init`.

I think the call_spiders function is a good idea in this case, to do a cast to self to refresh the list. Then perhaps the `Crawly.list_spiders` function should be consolidated into this function as well.

oltarasenko:
No no, I was not talking about Registry. I was suggesting to follow the approach of keeping the list in the Engine's state on init for example. And yes, I it should be unified with the one we have in top-level module..

Ziinc:
I see. I was thinking that the `Crawly` functions would be more appropriate for REPL usage (e.g. listing spiders if user forgot the name), which is why i placed the management functions in the engine, since it contain the spider's status as well, instead of just listing module names. 


for the `Crawly.Engine`, I'm thinking that we merge the `running_spiders` state key into a `spiders` key, such that the `:spiders` key holds all current spider info (current `job_id`, crawl stats, pid)
Then  `running_spiders` would just be a matter of doing an `Enum.filter` based on the `spiders` key. 

Another note, I'll rename the `state` to `status` on the spider info map , as it might give rise to confusion (since state can refer to so many things).

Ziinc:
Refactored the module.
changes:
- stores spiders on the :spiders key. 
- all calls will do a check if the state should be refreshed. Its when engine functions are called without first being initialized.

a breaking change is that running spiders returns a list instead of the previous map

I've also added a get_spider/1 function, which returns the spider_info type.

Ziinc:
I also had to move the spider in the MangerTest module to the compiled test_helper.ex file, as spiders defined in exs were not getting picked up by the refresh function

lib/crawly/engine.ex
```diff
@@ -33,58 +40,149 @@ defmodule Crawly.Engine do
...
+  end
+
   def start_link() do
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
   end
 
   @spec init(any) :: {:ok, __MODULE__.t()}
   def init(_args) do
-    {:ok, %Crawly.Engine{}}
+    spiders = do_refresh_spiders()
```
oltarasenko, line 68: I would suggest improving the naming. Maybe `get_spiders_list()`

lib/crawly/engine.ex
```diff
@@ -33,58 +40,149 @@ defmodule Crawly.Engine do
...
+  end
+
+  def handle_call(:list_spiders, _from, state) do
+    state = maybe_refresh_state(state)
+
+    {:reply, state.spiders, state}
   end
 
   def handle_call({:start_spider, spider_name}, _form, state) do
+    state = maybe_refresh_state(state)
```
oltarasenko, line 96: Let's not refresh the list of spiders implicitly, it makes the code more complex.

lib/crawly/engine.ex
```diff
@@ -33,58 +40,149 @@ defmodule Crawly.Engine do
...
+  def list_spiders() do
+    GenServer.call(__MODULE__, :list_spiders)
+  end
+
+  @spec running_spiders() :: [spider_info()]
   def running_spiders() do
     GenServer.call(__MODULE__, :running_spiders)
   end
 
+  @spec get_spider(module()) :: spider_info()
```
oltarasenko, line 53: Should it be re-named to `get_spider_info` instead? What is the plan for this function? Do we want to get spider stats from it in the future?

lib/crawly/engine.ex
```diff
@@ -33,58 +40,149 @@ defmodule Crawly.Engine do
...
+  def get_spider(name) do
+    GenServer.call(__MODULE__, {:get_spider, name})
+  end
+
+  def refresh_spiders() do
```
oltarasenko, line 58: Maybe we can rename it to `refresh_spiders_list`

Ziinc:
leaving the internal state unchanged means maintaining the map with running spiders and the separate list of spiders and their status. Risks going out of sync and is tech debt to leave it in.

Ziinc:
will split up the PR

Ziinc:
I undid the internal api changes. the spider_info list is now created dynamically.

I also left in the `started_spiders` map, which is used to check if spiders are running.

I also made the status type (started/stopped) to be atom based, so that it will be a simpler data structure for end user usage.

Also removed the implicit refreshing. The list now needs to be refreshed manually. It also refreshes once on genserver init.

oltarasenko (approved):
Looks good!

PR #106, opened 2020-05-18
Adds Crawly.Engine.stop_all_spiders/0

Ziinc:
For #37 , builds on #104


Ziinc:
Made all tests pass, the stop_all_spiders function just delegates to the existing `stop_spider/2` function.

I also shifted down the on_spider_closed_callback execution position to be within the genserver call handler, so that we only execute the callback if the spider was closed (and not if the spider was not running and was not "closed" by the engine).

Issue #52, opened 2020-01-14
Implement Crawly.crawl/2 command for fetching given page with given spider


oltarasenko:
It may be Crawly.crawl(url, spider_name) or Crawly.crawl_with(url, spider_name)

The idea is based on: 
https://github.com/oltarasenko/crawly/pull/51#issuecomment-573744130

We want to build a way to debug how spider would fetch a given page. Which links are going to be extracted? Which items are going to be fetched?

Ziinc:
Perhaps instead of having a completely separate function, we can extend the fetch function to have a :with option for passing in the spider?

PR #107, opened 2020-05-18
Improved Crawly.fetch/2 to accept fetching with a spider

Ziinc:
for #52, implements a `:with` option that allows the passing of the spider name.

Breaking changes includes moving of the `headers` and `options` parameters to the `opts` parameter, as keyword lists instead of positional args.

### DIFF
--- a/documentation/basic_concepts.md
+++ b/documentation/basic_concepts.md
@@ -37,6 +37,11 @@ In order to make a working web crawler, all the behaviour callbacks need to be i
  to prepare first requests on `init()`. Which might be useful if, for example, you
  want to pass a session cookie to the starting request. Note: `start_requests` are
  processed before start_urls.
+ ** This callback is going to be deprecated in favour of init/1. For now the backwords 
+ compatibility is kept with a help of macro which always generates `init/1`.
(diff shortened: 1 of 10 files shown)


Ziinc:
Updated, tests passes on my side, should be ok to merge.

oltarasenko:
Looks extremely good. The question is: "Can we move it outside of the main module? E.g. I want to shrink the crawly.ex"

@Ziinc could you fix the test, alternatively, I can pick it up if you don't have time right now.

Ziinc:
We could move to `Utils`, but Utils is also getting quite fat

oltarasenko:
Yeah. I have the same feelings. I was thinking of splitting parts related to Crawly from parts related to spiders into separate modules.

Maybe it could make these modules small. In any case, can be done later on separately. Could you rebase it, so I will check again.

oltarasenko:
@Ziinc I am about to cut off a new release. Could you please rebase this change, so I can include it as well?

Ziinc:
Done!

oltarasenko (approved):

PR #108, opened 2020-05-19
Tag crawl attempts

oltarasenko:
Engine now support and optional crawl_id argument. If crawl_id
is not provided it will be generated automatically.

### DIFF
--- a/lib/crawly/engine.ex
+++ b/lib/crawly/engine.ex
@@ -13,12 +13,12 @@ defmodule Crawly.Engine do
 
   defstruct started_spiders: %{}
 
-  @spec start_spider(module()) ::
+  @spec start_spider(module(), binary()) ::
           :ok
           | {:error, :spider_already_started}
           | {:error, :atom}
-  def start_spider(spider_name) do
-    GenServer.call(__MODULE__, {:start_spider, spider_name})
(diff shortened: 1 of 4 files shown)


oltarasenko:
> I think we should wait until the list_spiders PR is settled, as it would change how the crawl_id is stored and passed around.
Ok, we could have some sort of agreement here, as I need this change in order to initiate jobs from Crawly UI properly. Otherwise, I have to guess which items I am getting.

> I don't think passing the crawl_id everywhere as parameters is a good idea.
Yes, I think it should stay in the engine. Anyone who wants to get a tag will query the engine. That's the intention I have atm.

> I also think that the handle_call for :start_spider should be the one generating the id.
I am not sure. Could you explain why it might be needed? E.g. we can still control if we want to store the UUID (e.g. if the spider was not started etc). So I can't see races or anything like that, which will make handle_call a better place for ID generation. Please advise.

Ziinc:
I think that keeping it simple with a :crawl_id on the stored spider info would be good. I think the get_crawl_id should be changed to get_spider_info

Ziinc:
As for generating the uuid in the handle_call, it's more about  keep the code localised, as the module would be getting quite big soon

Ziinc:
will review again once #104 is done

PR #111, opened 2020-05-19
Request throttling improvements

oltarasenko:
Previousely a worker was able to send up to 200 requests per minute
we decided to limit it a way more. Now one worker can do only 4 requests
per minute.

Also improved tests (removed ugly sleeps)


test/manager_test.exs
```diff
@@ -29,7 +31,7 @@ defmodule ManagerTest do
```
Ziinc, line 31: The test name doesn't actually describe what is being tested here. Perhaps rename to "max request per minute is respected"

PR #114, opened 2020-06-20
Format sample config.exs on README

TORIFUKUKaiou:
There is an extra space.

I listened to your ElixirConf EU session.
I enjoyed it.
I tried the Quickstart.
Good! It works easy for me.

I guess that sample config.exs has an extra space on README.


oltarasenko:
Looks good! Thanks for the feedback! It would be great to see you using Crawly for your own commercial projects!

PR #116, opened 2020-07-02
Feature/benchmark

filipevarjao:
This PR adds the possibility to run a local benchmark for web scraping tool performance, the measurement collected is the number of requests and items per minute, memory usage, and the number of reductions on the spider process.

It starts a dummy local HTTP server, generating severals URLs in order to perform concurrent requests per domain


test/bench_test.exs
```diff
@@ -0,0 +1,39 @@
+defmodule BenchTest do
```
oltarasenko, line 1: Why do we need this test?

test/features/managet_test_spider.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Features.Manager.TestSpider do
```
oltarasenko, line 1: I like the idea of splitting this code as you're suggesting. But could we address it in a separate PR?

test/manager_test.exs
```diff
@@ -1,6 +1,8 @@
 defmodule ManagerTest do
```
oltarasenko, line 1: Maybe this one is unrelated to bench as well.

test/bench_test.exs
filipevarjao: to keep the test coverage up to 80% and to show it hat is possible to run the benchmark with a different spider

lib/crawly/bench.ex
```diff
@@ -55,9 +60,20 @@ defmodule Crawly.Bench do
         } | Current crawl speed is: #{num_requests_now} requests/sec"
       )
 
+      num_of_workers = DynamicSupervisor.count_children(name)[:workers]
+
+      if rem(retries, 10) == 0 and retries != 100 do
+        Logger.info("Adding 10 workers for #{name}")
+
+        Enum.map(1..10, fn _x ->
+          DynamicSupervisor.start_child(name, {Crawly.Worker, [name]})
```
oltarasenko, line 69: I would rather prefer to have it inside crawly. To do something like Crawly.Manager.add_worker/spider_name

lib/crawly/bench.ex
```diff
@@ -0,0 +1,82 @@
...
+         interval
+       ) do
+    Process.sleep(interval)
+
+    if Process.whereis(name) do
+      spiders = Engine.running_spiders()
+
+      {:stored_requests, req_count} = Crawly.RequestsStorage.stats(name)
+
+      {_, pid, :worker, _} =
```
oltarasenko, line 44: I would want to extend Managers API to get a current list of workers

lib/crawly/bench.ex
```diff
@@ -0,0 +1,82 @@
...
+        Supervisor.which_children(Map.get(spiders, name))
+        |> Enum.find(&({Crawly.Manager, _, :worker, [Crawly.Manager]} = &1))
+
+      {:info, info} = GenServer.call(pid, :collect_metrics)
```
oltarasenko, line 48: This could be an API of CrawlyManager as well. Lets extend it as a separate PR please

lib/crawly/bench/bench_router.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Crawly.Bench.BenchRouter do
```
oltarasenko, line 1: I wanted to check if it's possible (or if we should do it) to abstract it to a separate node? E.g. having an exs file, which is started separately from Crawly. So potentially it's possible to run it as a standalone process, to avoid any possible collision.

lib/crawly/bench.ex
filipevarjao: https://github.com/oltarasenko/crawly/pull/118

PR #119, opened 2020-07-26
[Draft] Use fake user agents as defaults

### LINKED ISSUE #112: Improve user agents database
A lot of my crawl depends on proper user-agent strings. It's a bit hard to supply user agents using a config as we're doing now. It would be good to have a database of user agents and to pick user agents from it. I am thinking of a standalone application with a simple interface, which then could be integrated into crawly.

We could get a database from:http://www.useragentstring.com/pages/api.php or any other service.

sreecodeslayer:
Opening an early Draft PR since am new to this codebase.
Please feel free to correct me if something looks out of place 😄 

Closes #112


oltarasenko:
Hey @sreecodeslayer I am on vacation right now away from the computer. Will return in one week. Will check the PR afterward.

sreecodeslayer:
> Hey @sreecodeslayer I am on vacation right now away from the computer. Will return in one week. Will check the PR afterward.

Sure. Am still experimenting a bit on where to place the implementations 😃

Ziinc:
I'm of the opinion that faker should be an optional dependency. Not everyone needs/wants fake user agents.

sreecodeslayer:
> I'm of the opinion that faker should be an optional dependency. Not everyone needs/wants fake user agents.

Of course! That's the plan, unfortunately I got swamped with my own work. So couldn't continue with the draft. I'll see if I chip in some time in coming days.

Ziinc:
I think that we should split this into two middlewares (as @oltarasenko [mentions](https://github.com/oltarasenko/crawly/issues/112#issuecomment-653808793)), one for simplly setting the UA string (as `UserAgent` works right now), and another that extends it with generating a fake UA, maybe called `FakeUserAgent`, that internally makes a call to `UserAgent` to set the string.

PR #121, opened 2020-09-01
[Fix] Do not close spider with :itemcount_timeout when closespider_timeout is set to :disabled

maiphuong-van:
I notice that the spider is closed with reason "..itemcount timeout achieved" after 1 minute even though I do not config `closespider_timeout` (disabled by default). Oleg points out that it could be because in `Crawly.Manager.maybe_stop_spider_by_timeout/3` when limit is an atom, it is always evaluated to `true` and stop. I added a condition that the `limit` should be an integer, else it should go to the :ok clause.

### DIFF
--- a/test/manager_test.exs
+++ b/test/manager_test.exs
@@ -4,7 +4,6 @@ defmodule ManagerTest do
   setup do
     Application.put_env(:crawly, :concurrent_requests_per_domain, 1)
     Application.put_env(:crawly, :closespider_itemcount, 10)
-    Application.put_env(:crawly, :concurrent_requests_per_domain, 1)
 
     :meck.expect(HTTPoison, :get, fn _, _, _ ->
       {:ok,
@@ -53,7 +52,6 @@ defmodule ManagerTest do
   end
 
   test "Closespider timeout is respected" do
-
     Process.register(self(), :spider_closed_callback_test)
 
     # Ignore closespider_itemcount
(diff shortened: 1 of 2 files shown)


test/manager_test.exs
```diff
@@ -91,6 +90,20 @@ defmodule ManagerTest do
     assert %{} == Crawly.Engine.running_spiders()
   end
 
+  @tag timeout: 61_000
+  test "spider does not close after 1 minute when closespider timeout is disabled" do
+    Application.put_env(:crawly, :closespider_timeout, :disabled)
```
oltarasenko, line 95: @maiphuong-van What happens if you also use:
Application.put_env(:crawly, :manager_operations_timeout, 1_000)?

maiphuong-van: I can try that 👍

maiphuong-van: It works! I will push the update :)

test/manager_test.exs
```diff
@@ -4,7 +4,6 @@ defmodule ManagerTest do
   setup do
     Application.put_env(:crawly, :concurrent_requests_per_domain, 1)
     Application.put_env(:crawly, :closespider_itemcount, 10)
-    Application.put_env(:crawly, :concurrent_requests_per_domain, 1)
```
maiphuong-van, line 7: Duplicate with line 5 so I removed it

oltarasenko (approved):

PR #125, opened 2020-10-06
Add support for start_requests in Crawly.Spider.init callback

oltarasenko:
It seems to be required for the case when you want to start crawling
prepared requests instead of raw urls.

This is must for login functionality


Ziinc:
Might need some struct pattern matching for validation. Maybe use a `with do` statement?

oltarasenko:
I am not sure about that yet :(. It's a part of the Manager's init. So it is a bit late to do the validation there. But on the other hand I think I should filter out start_requests which are not Crawly.Request, as a safety measure. However if a user provides wrong URLs or start_requests we have really nothing to do, but to stop a spider. Which will be a good hint after all!

PR #133, opened 2020-11-09
Extend domain filter with "share link" urls

oltarasenko:
Looks like urls like https://twitter.com?link=our_link was not
filtered out before. I have added filter for this and tests

### DIFF
--- a/lib/crawly/middlewares/domain_filter.ex
+++ b/lib/crawly/middlewares/domain_filter.ex
@@ -20,8 +20,9 @@ defmodule Crawly.Middlewares.DomainFilter do
   def run(request, state, _opts \\ []) do
     base_url = state.spider_name.base_url()
     parsed_url = URI.parse(request.url)
+    host = parsed_url.host
 
-    case String.contains?(base_url, parsed_url.host) do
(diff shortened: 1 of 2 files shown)


lib/crawly/middlewares/domain_filter.ex
```diff
@@ -19,8 +19,9 @@ defmodule Crawly.Middlewares.DomainFilter do
 
   def run(request, state, _opts \\ []) do
     base_url = state.spider_name.base_url()
+    parsed_url = URI.parse(request.url)
 
-    case String.contains?(request.url, base_url) do
+    case String.contains?(base_url, parsed_url.host) do
```
Ziinc, line 24: This will raise errors when `parsed_url.host` is `nil`, such as when the request url is relative. Perhaps a `nil` check would be good.

Not too sure about the expected behaviour for requests built with relative urls. I would think that it would skip the domain filter entirely.

Ziinc (approved):

lib/crawly/middlewares/domain_filter.ex
oltarasenko: Ok, I have added a check. Just in case.

PR #134, opened 2020-11-09
Use modern way of print a stacktrace from the worker

oltarasenko:


Ziinc (approved):
Looks ok, though we're double-logging the error variable, since it'll be shown in the formatted stacktrace too.

oltarasenko:
for some reason, the stacktrace was not visible. Probably the inspect on error, was swallowing the real stacktrace.

PR #137, opened 2020-11-12
Formatted the code

oltarasenko:


oltarasenko:
Just a `mix format`;

Ideally, we will spend more time on formatting concerns in the future

Issue #141, opened 2020-11-23
Announcement: Now it's possible to create a spider in UI


oltarasenko:
Hey people,

I want to make a short announcement in the inner circle, before moving forward in other places. I have just released an alpha version of UI based spider generators. You can find them here: http://crawlyui.com/spider/new

And here you can find an example of a spider created this way: http://crawlyui.com/spider/new?spider=ErlangSolutionsBlog

Would really appreciate feedback & ideas: @Ziinc

oltarasenko:
OK, I have managed to break everything...

oltarasenko:
Ok, it works again

Ziinc:
@oltarasenko I'm a little confused as to how the spiders are dynamically created. From skimming the `crawly_ui` source, I notice that there isn't a base template spider module being used. 

As mentioned [previously](https://github.com/oltarasenko/crawly/issues/126#issuecomment-708201573), managing dynamic spiders is definitely very useful, because managing and monitoring dozens of spiders manually is quite a hassle.

Could you give a brief explanation on how the spider is being dynamically instantiated and passed to the crawly engine? I understand that the extraction rules and init is stored in ecto, but what about the spider-level settings? and how does the engine keep track of it, considering that the spider discovery is based on module behaviour implementation?

oltarasenko:
Hey @Ziinc the feature is quite experimental atm, and some parts would have to be improved (as they look ugly). However I wanted to build this feature, and it feels like it's almost working. Once done, I will be able to re-write it, to make it look normal from the coding point of view. (Of course, if there will be interest in this feature after all. Actually, I've noticed that in general, the interest in scraping and crawling is quite limited in the Elixir community, so it may be the case that these efforts were not needed :().


Regarding the implementation:
1. Currently we have one spider template, which is defined as a string for now (🤮 ). And the idea is that a spider module is generated on the fly from the string + data from the database (https://github.com/oltarasenko/crawly_ui/blob/master/lib/crawly_ui.ex#L30). I want to make a working version here, and then I will try to find a better way..
2. Once the create_spider function is called, the module is generated from the string and dynamically loaded into a worker node, so the worker can pick it up
3. Finally once code is loaded the UI schedules it, similarly as it does it for other spiders.

Other questions:
1. For now spider level settings are also in the same module as the extraction rules. However, in the future, I plan to allow crawly_ui to schedule a spider with given settings (so settings will not be a part of a given spider anymore)
[...]

PR #148, opened 2020-12-10
Runtime spider creation

Ziinc:
This PR aims to allow for the easy creation of new spiders at runtime, utilizing modules that define the `Crawly.Spider` behaviour as spider templates.

Considered success if:
1. Can create multiple spiders at runtime from the engine using the same spider template.
2. Can stop the runtime spider from the engine.
3. Can scrape items normally with a runtime spider

Stretch:
- Can provide initialization options to the runtime spider, which can then be consumed by the spider template 

Initial discussion here: https://github.com/oltarasenko/crawly/issues/141#issuecomment-735245579

Main implementation notes:
- Use strings for spider names for referencing an instantiated spider 
- allow a template option to specify a spider module to be used as a template.


Ziinc:
@oltarasenko I can't replicate the circle ci tests failure on my local machine, for some reason. All tests are passing on my side. In any case, do have a look at the changes:

1. Engine starts and stops spiders asyncronously now, and I've refactored the starting/stopping to be asyncronously triggered so that the engine does not block when spider info is retrieved by other processes down the tree.
2. I've tried to consolidate the tests and the Engine/Manager APIs to make it more consistent, so some functions have been flagged for deprecation.
3. I've added 2 local registries for worker pool and manager name referencing, so that it is decoupled from the spider tempalte module. They're supervised in the application.ex supervision tree.  

I'll be adding in more docs and trying to hunt down the 3 failing tests. Let me know what you think

PR #149, opened 2020-12-10
Telemetry metrics support

Ziinc:
This PR introduces telemetry support to allow easy monitoring of crawling operations.

Considered success if:
1. Can monitor scraping-related events (requests made, redirections, response codes, dropped requests, dropped items, etc)
2. Can monitor spider-related events (requests stored, items stored, items parsed)
3. Can monitor engine-related events (spider started, spider stopped, worker/manager crashes)

Originally discussed here: https://github.com/oltarasenko/crawly/issues/141#issuecomment-735245579

### DIFF
--- a/lib/crawly/telemetry.ex
+++ b/lib/crawly/telemetry.ex
@@ -1,0 +1,38 @@
+defmodule Crawly.Telemetry do
+  @moduledoc """
+  Module for telemetry support for Crawly monitoring
+
+  To disable VM metrics, add this to yout config:
+  ```elixir
+    config :telemetry_poller, :default, false
+  ```
+  """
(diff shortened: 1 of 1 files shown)


oltarasenko:
@Ziinc just in case, how do we render the telemetry results?

Ziinc:
I think the rendering could either part of crawly_ui or as a separate package like crawly_live_metrics, like how phoenix does it. 
But instead of a full-fledge dashboard, i was thinking of widgets that can be simply plugged into any existing phoenix app. So then crawly_ui can make use of it too, without the metrics display being overly coupled with the entire crawly_ui application

PR #151, opened 2020-12-10
Bugfix for Manager timing out on high intial url/request count

Ziinc:
This is a bugfix where the Manager crashes due to timing out on the `init` callback, especially when there is a high number of start requests/urls.

This PR implements a split strategy for storing urls/requests using both sync and async methods, by storing the first 1000 requests and firing off a linked task that adds the remaining requests.

### DIFF
--- a/lib/crawly/manager.ex
+++ b/lib/crawly/manager.ex
@@ -66,28 +66,34 @@ defmodule Crawly.Manager do
     Process.link(request_storage_pid)
 
     # Store start urls
-    Enum.each(
-      Keyword.get(init, :start_requests, []),
-      fn
-        %Crawly.Request{} = request ->
-          Crawly.RequestsStorage.store(spider_name, request)
+    start_requests = Keyword.get(init, :start_requests, [])
+    {sync_start_req, async_start_req} = Enum.split(start_requests, 1000)
 
-        request ->
-          # We should not attempt to store something which is not a request
-          Logger.error(
-            "#{inspect(request)} does not seem to be a request. Ignoring."
-          )
+    store_req_fun = fn
+      %Crawly.Request{} = request ->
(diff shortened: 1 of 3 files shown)


oltarasenko:
Thanks for the PR @Ziinc. Indeed the case with bug number of start URLs is currently incompatible with Crawly.

However, as soon as we're going to fix it anyway I want to suggest the following:
1. We should simplify init. Long term operations in init may cause crashes depending on supervision strategies. So I was thinking about using handle_continue (or init_ack + enter_loop) for URLs initialization;
2. I think at this point we should also discuss a possibility to support incoming streams of data. As having 400_000 URLs list in memory does not sound quite right
3. Regarding the async Tasks - I would try to avoid having them in the Manager if possible, however, it might be a topic for further investigation.

oltarasenko (changes requested):
Thanks for the PR @Ziinc. Indeed the case with bug number of start URLs is currently incompatible with Crawly.

However, as soon as we're going to fix it anyway I want to suggest the following:

1. We should simplify init. Long term operations in init may cause crashes depending on supervision strategies. So I was thinking about using handle_continue (or init_ack + enter_loop) for URLs initialization;
2. I think at this point we should also discuss a possibility to support incoming streams of data. As having 400_000 URLs list in memory does not sound quite right
3. Regarding the async Tasks - I would try to avoid having them in the Manager if possible, however, I don't see how to achieve it now :(, it might be a topic for further investigation.

Ziinc:
OK. for (1), will refactor to utilize `handle_continue`.
For (3), since the start urls/requests are one-off insertions, I think placing them in the manager is fine. We add async bulk request insertion to the RequestStorage, if that is what you mean.

Will open up separate issues for (2) once this is merged.

PR #155, opened 2020-12-14
Split logs

oltarasenko:


Ziinc:
I think that passing around the spider name AND the craw_id seems unnecessary. I think passing around the crawl_id and using a lookup function to check the spiders stored in the engine would be a better approach. This way, the crawl_id functions as a primary key.

I think an ets table in the engine would allow us to reference the spider module much more easily, instead of passing the module name around. It would also make the runtime spiders PR easier to implement, by decoupling the running of spiders from the module.

Ziinc:
It might also be beneficial to wrap Logger with a utility function, so that we can standardize the way the spider metadata is logged

oltarasenko:
> I think that passing around the spider name AND the craw_id seems unnecessary. I think passing around the crawl_id and using a lookup function to check the spiders stored in the engine would be a better approach. This way, the crawl_id functions as a primary key.

I have the same feelings. At first, I did not want to do this full switch from spider_name to crawl_id, but now as soon as I have a bit of time I will do this exercise.

> I think obtaining the spider module as needed from the engine (whether as an ets table or from genserver state) in the engine would be much easier, instead of passing the module name around. It would also make the runtime spiders PR easier to 
implement, by decoupling the running of spiders from the module.

Yes, but I will start with something simple at first. Let me see.

> It might also be beneficial to wrap Logger with a utility function, so that we can standardize the way the spider metadata is logged

Looks like a better way to do it is to call `Logger.metadata(crawl_id: 1)` in related processes.

oltarasenko:
@Ziinc I have made another review of the code. Looks like the switch to the use of crawl_id will require quite a bit of work, which implies a change to how do we pick spiders or store things.

I also not quite sure what to do with Requests/DataStorage workers, which use a spider_name in order to avoid creating duplication of these processes.

I would rather defer this question to a separate unrelated discussion.

mix.exs
```diff
@@ -72,7 +72,8 @@ defmodule Crawly.Mixfile do
       {:ex_doc, "~> 0.19", only: :dev, runtime: false},
       {:earmark, "~> 1.2", only: :dev},
       {:meck, "~> 0.9", only: :test},
-      {:excoveralls, "~> 0.10", only: :test}
+      {:excoveralls, "~> 0.10", only: :test},
+      {:logger_file_backend, "~> 0.0.11"}
```
Ziinc, line 76: should be an opt-in dep.

test/manager_test.exs
```diff
@@ -31,7 +31,7 @@ defmodule ManagerTest do
 
   test "it is possible to add more workers to a spider" do
     spider_name = Manager.TestSpider
-    :ok = Crawly.Engine.start_spider(spider_name)
+    :ok = Crawly.Engine.start_spider(spider_name, crawl_id: "add_workers_test")
```
Ziinc, line 34: the engine should generate a crawl_id automatically. the change is unnecessary and isn't part of the test scope

test/middlewares/domain_filter_test.exs
```diff
@@ -16,7 +16,7 @@ defmodule Middlewares.DomainFilterTest do
   test "Filters out requests that do not contain a spider's base_url" do
     middlewares = [Crawly.Middlewares.DomainFilter]
     req = %Crawly.Request{url: "https://www.some_url.com"}
-    state = %{spider_name: :test_spider}
+    state = %{spider_name: :test_spider, crawl_id: "id"}
```
Ziinc, line 19: testing for whether the crawl_id is set in state by the request storage worker should be in the request storage worker test, not in the middlewares tests.

test/pipelines/json_encoder_test.exs
```diff
@@ -6,7 +6,7 @@ defmodule Pipelines.JSONEncoderTest do
   test "Converts a given map to a json string" do
     pipelines = [Crawly.Pipelines.JSONEncoder]
     item = @valid
-    state = %{}
+    state = %{spider_name: Test, crawl_id: "test"}
```
Ziinc, line 9: same as the middlewares, testing for crawl_id should be in the data storage worker test file

Ziinc:
i was reviewing it midway when you merged it haha.

wrt the passing of the crawl_id around, i'll revisit it when i work on the runtime spiders. Right now it feels like a lot of room for error.

oltarasenko:
Sorry for that :(. I will address the comments above on master :(.

Yeah, unfortunately, the switch to crawl_id looks a bit complex, however, I think we should do it.

PR #156, opened 2020-12-14
Trying to make tests more stable

oltarasenko:


oltarasenko:
I will merge it, as soon as I think it improves the situation with tests at least a bit. Will continue work on tests in next PRs

PR #161, opened 2020-12-21
Simplify manager init function

oltarasenko:
Hey @Ziinc I have simplified the handle_continue function a bit (e.g. combined operations for requests and URLs). Hopefully, that looks ok to you.

### DIFF
--- a/lib/crawly/manager.ex
+++ b/lib/crawly/manager.ex
@@ -28,6 +28,7 @@ defmodule Crawly.Manager do
   require Logger
 
   @timeout 60_000
+  @start_request_split_size 50
 
   use GenServer
 
@@ -66,6 +67,25 @@ defmodule Crawly.Manager do
 
     Process.link(request_storage_pid)
 
+    # Add start requests to the requests storage
+    init = spider_name.init(options)
+
+    start_requests_from_req = Keyword.get(init, :start_requests, [])
+
+    start_requests_from_urls =
+      init
+      |> Keyword.get(:start_urls, [])
+      |> Crawly.Utils.requests_from_urls()
+
+    start_requests = start_requests_from_req ++ start_requests_from_urls
+
+    # Split start requests, so it's possible to initialize a part of them in async
+    # manner
+    {start_req, async_start_req} =
+      Enum.split(start_requests, @start_request_split_size)
+
+    :ok = store_requests(spider_name, start_req)
+
     # Start workers
     num_workers =
       Utils.get_settings(:concurrent_requests_per_domain, spider_name, 4)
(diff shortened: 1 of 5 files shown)


lib/crawly/manager.ex
```diff
@@ -66,6 +67,25 @@ defmodule Crawly.Manager do
...
+      |> Crawly.Utils.requests_from_urls()
+
+    start_requests = start_requests_from_req ++ start_requests_from_urls
+
+    # Split start requests, so it's possible to initialize a part of them in async
+    # manner
+    {start_req, async_start_req} =
+      Enum.split(start_requests, @start_request_split_size)
+
+    :ok = store_requests(spider_name, start_req)
```
Ziinc, line 87: What are the benefits of adding in the synchronous requests at this stage? Does it solve a race condition for the tests?

IMO it would be clearer to have all request insertions in their own handle_continue block, as before, which makes it easier to locate.  

I do agree with the combining of requests/urls to avoid duplication.

lib/crawly/manager.ex
```diff
@@ -95,45 +115,15 @@ defmodule Crawly.Manager do
        tref: tref,
        prev_scraped_cnt: 0,
        workers: worker_pids
-     }, {:continue, {:startup, options}}}
+     }, {:continue, {:startup, {spider_name, async_start_req}}}}
```
Ziinc, line 118: Passing args to the handle_continue `:startup` in this manner limits future extensions. Seems like an unnecessary change.

lib/crawly/manager.ex
```diff
@@ -225,4 +215,13 @@ defmodule Crawly.Manager do
     do: String.to_integer(value)
 
   defp maybe_convert_to_integer(value) when is_integer(value), do: value
+
+  defp store_requests(spider_name, requests) do
```
Ziinc, line 219: maybe rename to `store_requests_sync`

Ziinc (approved):
Changes look fine, but a comments/improvements for code quality.

lib/crawly/manager.ex
oltarasenko: I was catching some strange cases with workers (in tests only, in general, I think it's a problem of tests which needs to be rewritten in some cases). So I want to feed the RequestsStorage as soon as possible. But I am getting your point, let me think a bit more about the case.

oltarasenko:
Another problem I am thinking of is that if we have a huge list of links it may be the case that we will block the requests storage with store operations. So it will be almost impossible to do the pop :(. Maybe we should add a small delay to the async operation, just in case, so it's possible to do something with workers?

lib/crawly/manager.ex
oltarasenko: I am not sure about it. Basically, I would only mark async functions, and assume everything else is sync.

oltarasenko:
@Ziinc I have addressed the issues you have pointed out. (Mainly moved all requests back to the handle_continue)

Ziinc (approved):
looks great :+1:

PR #162, opened 2020-12-22
[Draft] Add ui module retrival from config

### LINKED ISSUE #152: Configurable SendToUI Pipeline
It would be nice to be able to select the module and the function to which the data is sent when doing 
```
:rpc.cast(ui_node, CrawlyUI, :store_item,
```
 is it something you could be interested in?
For context:
I'm using crawly to scan amazon, on top of it I've added something similar to `CrawlyUI` and I would like to send the item to that module. I workaround it by having a module called CrawlyUI that can handle the call but I would be nice to add it in the config

FrancescoZ:
Add a new config to notify a different module when an Item is created.

The pipeline still using CrawlyUI as a default value if the configuration is not set

This resolve #152


FrancescoZ:
@oltarasenko Not sure how I should test this. I cannot us `meck` module to test the call, any suggestion?

lib/crawly/pipelines/experimental/send_to_ui.ex
```diff
@@ -15,6 +15,12 @@ defmodule Crawly.Pipelines.Experimental.SendToUI do
 
     spider_name = state.spider_name |> Atom.to_string()
 
+    ui_module =
```
Ziinc, line 18: You can make it more concise with `Keyword.get(opts, :ui_module, CrawlyUI)`

Ziinc (changes requested):
Add in a test for the `:rpc` call.

You can use `:meck` to mock the `:rpc` module, then check the args that was passed to verify that it was called with your given ui module.

Ziinc:
@FrancescoZ  is `:rpc` [one of the erlang modules that cannot be mocked](https://github.com/eproxus/meck#caveats)?

One way to test this (that i can think of) is to send it to itself, then have a mocked module passed as the ui_module where you can then verify that `store_item` was called.

oltarasenko:
Looks good to me. If it's possible to add a test that tries to send an rpc call to itself, it could be beneficial. (Otherwise, I would not try to mock RPC).

oltarasenko:
@FrancescoZ I still have a bit of development time in my timeslot. Should I take over this task? (In case if you're stuck with other things)

Ziinc (approved):

oltarasenko:
I have decided to abandon the CrawlyUI project for now. I don't have any spare time to maintain both projects, so I think this one is not a valid thing anyway.

PR #164, opened 2020-12-22
Add general purpose link extractor

oltarasenko:
Add a Utility function which simplifies links extraction, and allows
to get requests from a page by providing raw html and a list of
extraction patterns


Ziinc:
I'll do up the parsers PR first, so then efforts aren't duplicated. Ping you in a day or so.

oltarasenko:
Ok, I was not aware you're doing similar work already :)

oltarasenko:
@Ziinc I wonder if I can still merge this. E.g. it's a utility function we might still want to use. Otherwise, it's simple to get rid of it, once we will find a better way.

Ziinc:
Maybe you could refactor this into a `ExtractRequests` module? It could be the first parser that we implement by the next release.

oltarasenko:
Hey @Ziinc I have tried to approach the task with the parser's idea. At this moment code looks a bit complex to me, also looks like we need to add floki to deps in order to compile it.

What do you think about the implementation? Do I get the idea correctly?

Ziinc:
Would it be better to name it `ExtractRequests` instead? I fear that introducing the term "extractor" will make people think that it is yet another new concept to learn.

lib/crawly/utils.ex
```diff
@@ -174,6 +174,65 @@ defmodule Crawly.Utils do
...
+  performant comparing to the approach when links are extracted directly (e.g.
+  from the pagination, etc). So to summarize, it's slow, but in most cases that
+  will not be noticeable.
+  """
+  @spec extract_requests(page_body, base_url, list_of_filters) :: result
+        when page_body: String.t(),
+             base_url: String.t(),
+             list_of_filters: String.t(),
+             result: [Crawly.Request.t()]
+  def extract_requests(page_body, base_url, filters) do
```
Ziinc, line 196: I think that this entire function should be placed inside the `Crawly.Parsers.ExtractRequests` module, with the main logic inside the `run/3` and perhaps a convenience function for people to use it in `parse_items/1` if they want to.

Ziinc: The way i see the `Utils` module is for commonly used utility functions that are critically used within the Engine and also for Spiders and Pipelines, so that things conform to what the Engine/Manager/Worker expects.

lib/crawly/parsers/links_extractor.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Crawly.Parsers.RequestsExtractor do
+  @moduledoc """
+  Links extractor parser helper, which simplifies the process
+  of links extraction.
+  """
+
+  alias Crawly.Spider.Parse
+
+  def parse(parse_struct) do
+    %Parse{response: response, selector: selector} = parse_struct
```
Ziinc, line 10: was removed in main PR.

This module should  implement the `Crawly.Pipeline` behaviour, which will then have a `run/3` function that is then piped. The 2nd arg, state, will have a `:response` key and a `spider_name` key. Opts will contain the `:selector`

Ziinc (changes requested):
Hi i removed the `Parse` module in the main PR, and I forgot to remove the comments (which must have been a little confusing, because i wrote them when I was still prototyping it)

It should be completely identical to how middlewares/item pipelines are implemented.

oltarasenko:
I have decided to re-think the feature a bit. Will close it for now.

@Ziinc I am going to prepare 0.13.0 release today or tomorrow. Please let me know if you want to include something before that happens.

PR #165, opened 2020-12-24
Start with parameters

oltarasenko:


lib/crawly/worker.ex
```diff
@@ -9,6 +9,7 @@ defmodule Crawly.Worker do
 
   # define the default worker fetch interval.
   @default_backoff 10_000
+  @start_timeout 1000
```
oltarasenko, line 12: I had to tweak this value a bit, as workers were stuck a bit during the startup (as requests storage was empty), so in general, it caused delays due to increasing backoff.

Another aspect - it might be wise to start workers slowly (so we're not hitting the website at the same time), which probably can be achieved if we will add some randomness here.

Ziinc: perhaps a queue for worker requests?

test/manager_test.exs
```diff
@@ -84,124 +91,50 @@ defmodule ManagerTest do
...
-    assert_receive recv_opts
-    assert is_binary(recv_opts[:crawl_id])
-    assert Enum.sort(recv_opts[:urls]) == Enum.sort(urls)
+    Process.sleep(100)
+    assert_receive {:performing_request, req}
+    assert req.url == "https://www.example.com/blog.html"
   end
 end
 
 defmodule Manager.TestSpider do
```
Ziinc, line 118: I think instead of having a separate spider in the .exs test script, we should just mock the callbacks of the `TestSpider` module that is in the test_helpers.ex file, similar to how i did in the parse pipelines PR. it should make the tests less flaky.

Ziinc (approved):
Code looks fine

test/manager_test.exs
oltarasenko: I have created a new ticket for that

lib/crawly/worker.ex
oltarasenko: It would change the worker's logic quite a bit. I would return to this question separately.

PR #167, opened 2020-12-29
Improve on_spider_closed_callback

oltarasenko:
The collback is now called in the context of Engine gen server,
and only executed if spider is stopped.

Also extended a function signature, so now it takes crawl_id and
the reason.


Ziinc (approved):
If i'm not wrong, i had done this change in https://github.com/oltarasenko/crawly/pull/106.

Looks good

Ziinc:
actually, it would constitute a breaking change. Might be good idea to add that to the docs for when you bump the version

PR #170, opened 2021-01-30
File logging improvements

oshosanya:
- Allow opt-in for file logging, with false as default
- Added timestamp as prefix to log file name


lib/crawly/engine.ex
```diff
@@ -242,12 +242,17 @@ defmodule Crawly.Engine do
 
   defp set_spider_log(spider_name, crawl_id) do
     log_dir = Crawly.Utils.get_settings(:log_dir, spider_name, "/tmp")
```
Ziinc, line 244: could we switch the hardcoded `/tmp` to a generated path, so that it will be windows friendly 

https://til.hashrocket.com/posts/8rkl48vqmc-get-a-writable-temporary-directory-in-elixir

config/config.exs
```diff
@@ -45,6 +45,7 @@ config :crawly,
 
   # TODO: this looks outdated
   follow_redirect: true,
+  enable_file_logging: false,
```
Ziinc, line 48: Can we change it to `log_to_file`? so that it will be similar to the `log_dir` setting?

documentation/configuration.md
```diff
@@ -135,6 +135,12 @@ default: /tmp
 Set spider logs directory. All spiders have their own dedicated log file
 stored under the `log_dir` folder.
 
+### enable_file_logging :: String.t()
+
+default: false
+
+Allows to enable or disable file logging.
```
Ziinc, line 142: Enables or disables file logging.

documentation/configuration.md
```diff
@@ -135,6 +135,12 @@ default: /tmp
 Set spider logs directory. All spiders have their own dedicated log file
 stored under the `log_dir` folder.
```
Ziinc, line 136: Add in a reference to the `log_to_file` config. "Ignored if logging not enabled"

lib/crawly/engine.ex
```diff
@@ -242,12 +242,17 @@ defmodule Crawly.Engine do
...
-    Logger.add_backend({LoggerFileBackend, :debug})
+    enable_file_logging = Crawly.Utils.get_settings(:enable_file_logging)
+    current_unix_timestamp = :os.system_time(:second)
 
-    Logger.configure_backend({LoggerFileBackend, :debug},
-      path: "/#{log_dir}/#{spider_name}/#{crawl_id}.log",
-      level: :debug,
-      metadata_filter: [crawl_id: crawl_id]
-    )
+    if enable_file_logging do
```
Ziinc, line 248: to make it more concise, could you do:

```elixir
if Crawly.Utils.get_settings(:log_to_file) do
    log_dir = Crawly.Utils.get_settings(:log_dir, spider_name, "/tmp")
    current_unix_timestamp = :os.system_time(:second)
end
...

might be cleaner to move the log_to_file check to the main function call, removing the if-do

Ziinc: Additionally, should use get_settings/2 instead,  should include the spider_name, as we'll want to check if there are spider-level overrides.

Ziinc (changes requested):
could you also add in a test that tests for the enabling/disabling of the file logging?

nice work

lib/crawly/engine.ex
oshosanya: Done

config/config.exs
oshosanya: Done

documentation/configuration.md
oshosanya: Done

documentation/configuration.md
oshosanya: Done

lib/crawly/engine.ex
oshosanya: Done

Ziinc (approved):
I've added necessary tests and refined the path building.

Ziinc:
@oltarasenko will merge this in when checks are done, then you can go ahead and cut the v0.13 release

PR #177, opened 2021-03-19
fix some test error & compile time warning

feng19:
* fix test `list_known_spiders/0` error
* implement `get_spider_info/1` handle_call missing
* fix some `mix compile` & `mix test` compile time warning (type/docs/deprecated...)
* remove 'Elixir.' prefix module's format when logging
* enhancement Exception log by use `Exception.format/3`
* skip test "manager does not crash with high number of urls" because Crawly.RequestsStorage is a bottleneck


oltarasenko (approved):
Good work @feng19 . Thanks for finding time and contributing!

PR #178, opened 2021-03-24
Use keyword list for options

spectator:
I'm still new to Elixir so I might be missing something, but it looks like options were meant to be keyword list unlike what typespec says.

For example here is splash uses options as keyword list https://github.com/oltarasenko/crawly/blob/master/lib/crawly/fetchers/splash.ex#L32


Ziinc (approved):
thanks for the report :+1:

PR #180, opened 2021-04-10
Pluggable fetchers improvement, closes #43

### LINKED ISSUE #43: Pluggable fetchers improvements
- [ ] Update Crawly.fetch to use backends
- [ ] Allow specifying fetcher without options

oshosanya:


lib/crawly/worker.ex
```diff
@@ -76,16 +76,26 @@ defmodule Crawly.Worker do
              spider_name: atom(),
              response: HTTPoison.Response.t(),
              result: {:ok, {response, spider_name}} | {:error, term()}
-  defp get_response({request, spider_name}) do
+  def get_response({request, spider_name}) do
```
Ziinc, line 79: as this is an function being exposed for internal use only,  can you add a `@doc false` so that it will not be shown in docs.

lib/crawly/worker.ex
```diff
@@ -116,7 +126,7 @@ defmodule Crawly.Worker do
              parsed_item: Crawly.ParsedItem.t(),
              next: {parsed_item, response, spider_name},
              result: {:ok, next} | {:error, term()}
-  defp parse_item({response, spider_name}) do
+  def parse_item({response, spider_name}) do
```
Ziinc, line 129: same, should hide from docs

lib/crawly/worker.ex
```diff
@@ -76,16 +76,26 @@ defmodule Crawly.Worker do
...
-      )
+      case Crawly.Utils.get_settings(
+             :fetcher,
+             spider_name,
+             {Crawly.Fetchers.HTTPoisonFetcher, []}
+           ) do
+        {module, args} ->
+          {module, args}
+
+        {module} ->
```
Ziinc, line 93: this format is not supported, should not handle it. should throw an error if user attempts to use this format.

Ziinc: please remove this case.

lib/crawly/worker.ex
```diff
@@ -76,16 +76,26 @@ defmodule Crawly.Worker do
...
-        :fetcher,
-        spider_name,
-        {Crawly.Fetchers.HTTPoisonFetcher, []}
-      )
+      case Crawly.Utils.get_settings(
+             :fetcher,
+             spider_name,
+             {Crawly.Fetchers.HTTPoisonFetcher, []}
+           ) do
+        {module, args} ->
```
Ziinc, line 90: can you add in an `is_keyword` guard to ensure that args is a keyword list.

lib/crawly/worker.ex
```diff
@@ -76,16 +76,26 @@ defmodule Crawly.Worker do
...
+          {module, args}
+
+        {module} ->
+          {module, nil}
+
+        module ->
```
Ziinc, line 96: also, add in an `is_atom` guard for the module.

as a side note, this unwrapping logic can probably be moved to a utils function, as it is used by other workers as well.

Ziinc: also, can you add a default case which raises an invalid format error

Ziinc (changes requested):
refactor looks good, just some minor adjustments

Ziinc:
also, please rebase on master to clean up the commit history :+1:

lib/crawly/worker.ex
oshosanya: There's no `is_keyword` guard, checked around, could you send a pointer?

Ziinc: Ah my bad, you can use is_list instead

lib/crawly/worker.ex
oshosanya: Done

lib/crawly/worker.ex
oshosanya: Done

lib/crawly/worker.ex
oshosanya: Done

lib/crawly/worker.ex
oshosanya: Done

lib/crawly/worker.ex
oshosanya: Done

oshosanya:
> also, please rebase on master to clean up the commit history 👍

Done

lib/crawly/utils.ex
```diff
@@ -224,4 +224,27 @@ defmodule Crawly.Utils do
   def ensure_loaded?(module) do
     Code.ensure_loaded?(module)
   end
+
+  def unwrap_module_and_options(
```
Ziinc, line 228: didn't spot this before. Could you adjust this function so that it only unwraps a provided arg? So that we can just pipe the result of `Utils.get_settings` into the function: 
```elixir
Utils.get_settings(:fetcher)
|> Utils.unwrap_module_and_options()
```
also can you add a docs and example usage

lib/crawly/utils.ex
```diff
@@ -224,4 +224,27 @@ defmodule Crawly.Utils do
...
+           spider_name,
+           default_value
+         ) do
+      {module, args} when is_list(args) and is_atom(module) ->
+        {module, args}
+
+      module when is_atom(module) ->
+        {module, nil}
+
+      {module} ->
```
Ziinc, line 244: switch this to a `_` so that it'll match on anything.

lib/crawly/utils.ex
Ziinc: and also add typespecs. Function should take a term as first arg.

lib/crawly/utils.ex
```diff
@@ -224,4 +224,27 @@ defmodule Crawly.Utils do
...
+        raise "Invalid format: A #{setting} setting cannot be defined in the form `{#{
```
Ziinc, line 245: use `inspect()` to convert terms to string, and also add in the received input to the error string (for easier debugging)

lib/crawly/utils.ex
Ziinc: as to why this would be better, it reduces the interface of the function and aligns the function name with what it actually does.

Ziinc (changes requested):
If you've got the time, could you also add in the usage of `unwrap_module_and_options` to the data storage and request workers? currently they both implement their own unwrapping logic, it'll be good to standardize

lib/crawly/utils.ex
oshosanya: used `x` so I can use the value in `inspect/1`

lib/crawly/utils.ex
oshosanya: Done

lib/crawly/utils.ex
oshosanya: done

lib/crawly/utils.ex
```diff
@@ -224,4 +224,24 @@ defmodule Crawly.Utils do
...
+  """
+  @spec unwrap_module_and_options(term) ::
+          {atom, nil | maybe_improper_list}
+  def unwrap_module_and_options(setting) do
+    case setting do
+      {module, args} when is_list(args) and is_atom(module) ->
+        {module, args}
+
+      module when is_atom(module) ->
+        {module, nil}
```
Ziinc, line 239: should be `{module, []}`, options should always be a list.

Ziinc (changes requested):
just one comment on the default options value when unrapping, the rest looks great.

Ziinc (approved):
👌 👍

PR #191, opened 2021-08-17
Update example tutorial

oshosanya:
Updated the example in the tutorial to use `https://books.toscrape.com` as that website is less likely to change compared to homebase which is owned by an actual company.


documentation/tutorial.md
```diff
@@ -109,6 +109,49 @@ some functions:
...
+    [
+      start_urls: product_categories_urls
+    ]
+  end
+```
+
+We also need to add the `build_absolute_url/1` function which is used to build an absolute url from relative urls
+
+```elixir
+  defp build_absolute_url(url) do
```
Ziinc, line 146: This function is already available in `Crawly.Utils`

documentation/tutorial.md
```diff
@@ -109,6 +109,49 @@ some functions:
...
+As it is, the init function returns an empty array for `start_urls` which we will
+fix now. If you already know what urls, you want to start crawling, you can place them in
+the array, else you can pick the urls you want to start with from a particular page
+on your website of interest, in this case `https://books.toscrape.com`.
+
+Go ahead and modify your `init` to look this way.
+
+```elixir
+  @impl Crawly.Spider
+  def init() do
```
Ziinc, line 121: the init function should just provide the starting url and should be simpler 

(this is a quickstart afterall)

Keep the document parsing logic under `parse_item`

documentation/tutorial.md
```diff
@@ -160,92 +203,100 @@ selectors in Crawly shell.
...
+  end)
 ```
 
-Extract the `title` with:
+Here we select all products using the class name, iterate through it and pick the title using `Floki.find(x, "h3 a") |> Floki.attribute("title")`
+and pick the price using `Floki.find(x, ".product_price .price_color") |> Floki.text()`. The full text of the book name is only contained
+in the title attribute and not in the text node of the tag, hence our use of the title attribute.
+
+Also, the data is paginated and we are not only interested in the first page, we have to tell crawly to also visit the next page
+by creating a request struct with the url of the next page and other necessary data.
```
Ziinc, line 264: To handle pagination, we'll tell crawly to visit the next page by extracting the next pages' urls and building new requests from them.

documentation/tutorial.md
```diff
@@ -160,92 +203,100 @@ selectors in Crawly shell.
...
 
 ```
-{:ok, document} = Floki.parse_document(response.body)
-document |> Floki.find(".page-title h1") |> Floki.text()
-"Bosch UniversalImpact 800 Corded Hammer Drill"
+next_url =
```
Ziinc, line 267: request building seems a little verbose, and assumes only one paginator.

Floki.attribute should returna  list of strings. can map over it to create the requests.

dont need the case.

documentation/tutorial.md
```diff
@@ -160,92 +203,100 @@ selectors in Crawly shell.
...
+        build_absolute_url(response.request.url, url)
+        |> Crawly.Utils.request_from_url()
+
+      [new_request]
+  end
 ```
 
-Extract the `SKU` with:
+Here we get the url for the next page using Floki, which could be an empty string if such doesn't exist. We then create a request struct
+by piping the url we got from Floki to `build_absolute_url/2` and `Crawly.Utils.request_from_url`.
```
Ziinc, line 288: update to explain that if it doesn't exist (e.g. end of page) , it will simply return an empty list, and result in spider stopping.

Don't need to specially handle situation where only one exists etc.

documentation/tutorial.md
```diff
@@ -160,92 +203,100 @@ selectors in Crawly shell.
...
 
-```
-document |> Floki.find(".product-header-heading span") |> Floki.text
-"SKU:  494894"
-```
+Notice that we have not declared the function `build_absolute_url/2` yet, what we had earlier was `build_absolute_url/1`. This new function is necessary
```
Ziinc, line 290: Use the utils function

documentation/tutorial.md
```diff
@@ -160,92 +203,100 @@ selectors in Crawly shell.
...
 
-Extract the `price` with:
+We can now go ahead and add the `build_absolute_url/2` function to our code.
 
-```
-document |> Floki.find(".price-value [itemprop=priceCurrency]") |> Floki.text
-"£82"
+```elixir
+## Build absolute url supplying a different url as base
+defp build_absolute_url(base_url, url) do
```
Ziinc, line 297: utils function already has it

documentation/tutorial.md
```diff
@@ -255,18 +306,31 @@ just makes an `empty run`. Let’s integrate the extraction logic above
...
 
   @impl Crawly.Spider
   def init() do
+    # Get all the cateory URLs to use as our starting point
+    response = Crawly.fetch("https://books.toscrape.com/catalogue/category/books_1/index.html")
+
+    {:ok, document} = Floki.parse_document(response.body)
+
+    # Extract product category URLs
+    product_categories_urls =
```
Ziinc, line 323: category parsing can be in the parse_item, 

parse_item is flexible enough for multi pages structure extraction, dont need to specially have these in the init.

plus having it in the init makes it more complicated and may cause the spider to crash on error

documentation/tutorial.md
```diff
@@ -275,48 +339,53 @@ defmodule Homebase do
...
+    %Crawly.ParsedItem{:items => items, :requests => next_request}
   end
 
-  defp build_absolute_url(url), do: URI.merge(base_url(), url) |> to_string()
+  defp build_absolute_url(url) do
+    URI.merge(base_url(), url) |> to_string()
+  end
+
+  ## Build absolute url supplying a different url as base
+  defp build_absolute_url(base_url, url) do
```
Ziinc, line 386: see above.

documentation/tutorial.md
```diff
@@ -275,48 +339,53 @@ defmodule Homebase do
...
-      price:
-        document
-        |> Floki.find(".price-value [itemprop=priceCurrency]")
-        |> Floki.text()
-    }
+          [new_request]
+      end
 
-    %Crawly.ParsedItem{:items => [item], :requests => requests}
+    %Crawly.ParsedItem{:items => items, :requests => next_request}
```
Ziinc, line 378: use the normal atom syntax
%{items: ... ,, requests: ...}

documentation/tutorial.md
```diff
@@ -275,48 +339,53 @@ defmodule Homebase do
...
+
+    next_url =
       document
-      |> Floki.find("a.product-tile  ")
+      |> Floki.find(".next a")
       |> Floki.attribute("href")
+      |> Floki.text()
 
-    urls = product_pages ++ product_categories
+    next_request =
```
Ziinc, line 365: see above

Ziinc (changes requested):

documentation/tutorial.md
oshosanya: So there's product categories on each product category page, that would mean every time parse item runs, it'll attempt to add those urls again, although, it wouldn't go through. Just thought there wasn't any need to deliberately attempt to add the same urls multiple times and relying on the fact that they'll be parsed only once.

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
```diff
@@ -109,6 +109,14 @@ some functions:
 3. parse_item(): function which will be called to handle response
    downloaded by Crawly. It must return the `Crawly.ParsedItem` structure.
 
+As it is, the init function returns an empty array for `start_urls` which we will
+fix now. If you already know what urls, you want to start crawling, you can place them in
+the array, else you can pick the urls you want to start with from a particular page
+on your website of interest.
+
```
Ziinc, line 116: > The `init/0` function returns initialization parameters, with `:start_urls` set to an empty list by default. Let's update that to the url that we want our spider to begin crawling from. In this case, we set it to the homepage of `books.toscrape.com`.

Ziinc: Since it is a tutorial, we want to avoid readers from deviating from the example, so that they get expected results. Also, the concept of array slightly different for elixir, so better to stick with the correct terminology to avoid confusion.

documentation/tutorial.md
```diff
@@ -109,6 +109,14 @@ some functions:
...
+Using the shell, you can try selecting elements using Floki with the response. That gives you a faster way to test out your selectors
```
Ziinc, line 117: Mark this as a sidenote `using `>` blockquotes.

> Note: You can use the `iex` shell to test your selectors on the response HTML using Floki. Alternatively, use your browser's developer tools to test your selector.

documentation/tutorial.md
```diff
@@ -160,93 +168,77 @@ selectors in Crawly shell.
...
   status_code: 200
 }
 ```
 
-Using the shell, you can try selecting elements using Floki with the
-response. Let's say that we want to extract all product categories links from the
-page above:
+Now let's start parsing the document and extract the books in it.
 
 ```
```
Ziinc, line 204: add in syntax highlightiing `elixir`

documentation/tutorial.md
```diff
@@ -160,93 +168,77 @@ selectors in Crawly shell.
 1. Start the Elixir shell using `iex -S mix` command.
 2. Now you can fetch a given HTTP response using the following
    command:
-   `response = Crawly.fetch("https://www.homebase.co.uk/our-range/tools")`
+   `response = Crawly.fetch("https://books.toscrape.com/")`
 
 You will see something like:
 
 ```
```
Ziinc, line 175: syntax highlighting

documentation/tutorial.md
```diff
@@ -160,93 +168,77 @@ selectors in Crawly shell.
...
+  |> Enum.map(fn x ->
+    %{
+      title: Floki.find(x, "h3 a") |> Floki.attribute("title") |> Floki.text(),
+      price: Floki.find(x, ".product_price .price_color") |> Floki.text()
+    }
+  end)
 ```
 
-Extract the `title` with:
+Here we select all products using the class name, iterate through it and pick the title using `Floki.find(x, "h3 a") |> Floki.attribute("title") |> Floki.text()`
```
Ziinc, line 223: using the class `.product_pod`, then we iterate over each html fragment found and extract out relevant product information. In this case, we want to extract the title and price.

documentation/tutorial.md
```diff
@@ -160,93 +168,77 @@ selectors in Crawly shell.
...
+in the title attribute and not in the text node of the `a` tag, hence our use of the title attribute.
 
-```
-{:ok, document} = Floki.parse_document(response.body)
-document |> Floki.find(".page-title h1") |> Floki.text()
-"Bosch UniversalImpact 800 Corded Hammer Drill"
-```
-
-Extract the `SKU` with:
+To handle pagination, we'll tell crawly to visit the next page by extracting the next pages' urls and building new requests from them.
```
Ziinc, line 227: we'll tell Crawly

Refer to Crawly with capitalized C.

Add in the following note as well:

> Pagination url extraction can be within the same callback, as if there are no pagination HTML nodes found, `next_requests` would be an empty list.

documentation/tutorial.md
```diff
@@ -160,93 +168,77 @@ selectors in Crawly shell.
...
+  end)
 ```
 
-Extract the `price` with:
-
-```
-document |> Floki.find(".price-value [itemprop=priceCurrency]") |> Floki.text
-"£82"
-```
+Here we get the url for the next page using Floki, which could be an empty string if such doesn't exist, this would mean that we have gotten to the last page and will result in the spider stopping.
```
Ziinc, line 240: would it really result in an empty string? `Floki.attribute` should return an empty list if it does not find any nodes with `href`s

documentation/tutorial.md
```diff
@@ -160,93 +168,77 @@ selectors in Crawly shell.
...
 ```
 
-Extract the `price` with:
-
-```
-document |> Floki.find(".price-value [itemprop=priceCurrency]") |> Floki.text
-"£82"
-```
+Here we get the url for the next page using Floki, which could be an empty string if such doesn't exist, this would mean that we have gotten to the last page and will result in the spider stopping. 
+We then create a request struct by piping the url we got from Floki to `Crawly.Utils.build_absolute_url/2` and `Crawly.Utils.request_from_url/1`.
```
Ziinc, line 241: a `Crawly.Request` struct

Ziinc (changes requested):
Looks good, just needs some slight wording changes

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

documentation/tutorial.md
oshosanya: Done

Ziinc (approved):

Ziinc:
@oltarasenko will need you to merge in

PR #193, opened 2021-09-20
Fix Storage Worker exception

gfrancischelli:
Starting the same worker twice in a row was corrupting the `Crawly.RequestsStorage` GenServer state.

### DIFF
--- a/test/engine_test.exs
+++ b/test/engine_test.exs
@@ -14,6 +14,11 @@ defmodule EngineTest do
     end)
   end
 
+  test "starting a running spider returns an error" do
+    Crawly.Engine.start_spider(TestSpider)
+    assert {:error, :spider_already_started} Crawly.Engine.start_spider(TestSpider) 
+  end
+
   test "list_known_spiders/0 lists all spiders and their current status in the engine" do
     Crawly.Engine.refresh_spider_list()
     spiders = Crawly.Engine.list_known_spiders()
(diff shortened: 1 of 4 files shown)


test/request_storage_test.exs
```diff
@@ -23,6 +23,19 @@ defmodule RequestStorageTest do
     {:ok, %{crawler: :test_spider}}
   end
 
+  test "Starting a worker twice doesn't break", context do
```
Ziinc, line 26: could you adjust this test name to describe the expected underlying behaviour? "doesn't break" isn't very informative of what is being tested here.

test/engine_test.exs
```diff
@@ -14,6 +14,11 @@ defmodule EngineTest do
     end)
   end
 
+  test "starting a running spider returns an error" do
+    Crawly.Engine.start_spider(TestSpider)
+    assert {:error, :spider_already_started} Crawly.Engine.start_spider(TestSpider)
```
Ziinc, line 19: missing `=`

test/request_storage_test.exs
```diff
@@ -23,6 +23,19 @@ defmodule RequestStorageTest do
...
+    Crawly.RequestsStorage.start_worker(:test_spider, "crawl_id")
+
+    request = %Crawly.Request{
+      url: "http://example.com",
+      headers: [],
+      options: []
+    }
+
+    assert :ok = Crawly.RequestsStorage.store(context.crawler, request)
```
Ziinc, line 35: can we abstract these assertions to a `worker_is_healthy()` helper or something like that? seems repetitive to copy-paste them

Ziinc (changes requested):

PR #216, opened 2022-09-12
Make WriteToFile pipeline more robust

fstp:
Currently the pipeline will try to write given items to file but if the items do not implement String.Chars it will crash. By using Kernel.inspect any term can be saved without having to explicitly implement the protocol.

This was discovered in an example where the pipeline tried to write Crawly.Request structs but failed due to the struct not implementing String.Chars.


fstp:
This seems to be the wrong approach, will close this one and make a separate pull request where I implement the String.Chars protocol for the particular type instead.

PR #217, opened 2022-09-12
Implement String.Chars for Crawly.Request

fstp:
To enable the requests to be written to file/log (for example in the WriteToFile pipeline) the String.Chars protocol is implemented as a trivial Kernel.inspect dump.


Ziinc:
Hi thanks for the contribution, could you add in a test for this so that I can verify on my side the error that you mentioned in the closed PR?

fstp:
Sorry for the late response, sure I will add a testcase for it :)

PR #220, opened 2022-09-14
Adding the original request to the parse_item callback

nuno84:
Based on the discussion I opened: [discussion](https://github.com/elixir-crawly/crawly/discussions/215)
So, I want to crawl, lets say 100 websites and set the item parse elements (floki query elements) set on a webpage for the user to fine-tune. I found this Oleg article usefull: [article](https://oltarasenko.medium.com/using-elixir-and-crawly-for-price-monitoring-7364d345fc64) 
So my idea is to write a generic HTTP spider and inside it, use that info to parse the data based on the request custom_data.
I think this pull is usefull. And I tried to make it backward compatible.
So now you can do both:

`def parse_item(response), do: ...`
OR
`def parse_item(response, request = %Crawly.Request{custom_data: req_data}), do: ...`

The apply function is now:

```
  defp do_parse(nil, spider_name, response, request) do
    if :erlang.function_exported(spider_name, :parse_item, 2) do
      spider_name.parse_item(response, request)
    else
      spider_name.parse_item(response) # This is for backward compatibility
    end
  end
```

The tests are passing on my computer.
Please take a look.
I can add some documentation if this idea goes forward.
Thank you


starcraft66:
Super cool to see this, I was just about to go implement this myself after having a use-case for it (crawling image files for which the metadata is located on the previously-crawled page). I will see how well this works for me and take a look at the failing tests too. Unfortunate that development seems to be stalled on this project.

Ziinc:
@nuno84 PR is appreciated. The entire `Request` gets copied to the `Response` struct, so adding in the second argument to `parse_item` callback is unnecessary. 

I would suggest using the `metadata` or `meta` key on the `Request` struct instead of `custom_data`, which is more semantically correct.

lib/crawly.ex
```diff
@@ -79,8 +79,8 @@ defmodule Crawly do
 
       _ ->
         # spider provided, send response through  parse_item callback, pipe through the pipelines
-        with {:ok, {parsed_result, _, _}} <-
-               Crawly.Worker.parse_item({response, opts[:with]}),
+        with {:ok, {parsed_result, _, _, _}} <-
+               Crawly.Worker.parse_item({response, opts[:with], request}),
```
Ziinc, line 83: see comment

lib/crawly/request.ex
```diff
@@ -12,7 +12,8 @@ defmodule Crawly.Request do
             prev_response: nil,
             options: [],
             middlewares: [],
-            retries: 0
+            retries: 0,
+            custom_data: %{}
```
Ziinc, line 16: `metadata`

lib/crawly/request.ex
```diff
@@ -28,7 +29,8 @@ defmodule Crawly.Request do
           prev_response: %{},
           options: [option()],
           middlewares: [atom()],
-          retries: non_neg_integer()
+          retries: non_neg_integer(),
+          custom_data: %{}
```
Ziinc, line 33: `metadata`

lib/crawly/spider.ex
```diff
@@ -27,6 +27,10 @@ defmodule Crawly.Spider do
 
   @callback parse_item(response :: HTTPoison.Response.t()) ::
               Crawly.ParsedItem.t()
+  @callback parse_item(
+              response :: HTTPoison.Response.t(),
+              request :: Crawly.Request.t()
```
Ziinc, line 32: please rm

lib/crawly/worker.ex
```diff
@@ -136,19 +132,25 @@ defmodule Crawly.Worker do
     end
   end
 
-  defp do_parse(nil, spider_name, response),
-    do: spider_name.parse_item(response)
+  defp do_parse(nil, spider_name, response, request) do
```
Ziinc, line 135: see comment

lib/crawly/worker.ex
```diff
@@ -158,20 +160,22 @@ defmodule Crawly.Worker do
...
+  defp process_parsed_item({parsed_item, response, spider_name, _request}) do
     requests = Map.get(parsed_item, :requests, [])
     items = Map.get(parsed_item, :items, [])
     # Process all requests one by one
     Enum.each(
       requests,
-      fn request ->
-        request = Map.put(request, :prev_response, response)
-        Crawly.RequestsStorage.store(spider_name, request)
+      fn req ->
```
Ziinc, line 176: variable name change unnecessary, fully qualified names preferred for consistency.

Ziinc (changes requested):
please also perform formatting and add in docs on `metadata` usage for creating custom `Request` structs, and accessing them from the `parse_item/1` callback from the `request.metadata` path.

starcraft66:
> The entire Request gets copied to the Response struct

@Ziinc The `HTTPoison.Response.t()` contains the original `HTTPoison.Request.t()` struct but I don't think it is helpful in the context of the PR because that `HTTPoison.Request.t()` will not contain any of the metadata stored in the `Crawly.Request.t()` wrapping the response.

PR #248, opened 2023-04-02
Introduce SameDomainFilter

serpent213:
My application requires only one or two different Spiders but these are run over many different domains (passed as option to start_spider). So setting the base_url turned out to be difficult and I came up with this solution.

Thoughts?

### DIFF
--- a/documentation/basic_concepts.md
+++ b/documentation/basic_concepts.md
@@ -44,7 +44,7 @@ compatibility is kept with a help of macro which always generates `init/1`.
 `init(options)` same as `init/0` but also takes options (which can be passed from the engine during
 the spider start).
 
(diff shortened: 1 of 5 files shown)


oltarasenko:
@serpent213 The code looks quite nice. It would be great if you could write an article about your use case!

oltarasenko (approved):

serpent213:
> @serpent213 The code looks quite nice. It would be great if you could write an article about your use case!

You mean like a blog article? Good idea.

In one sentence, I'm building a general purpose search engine (think DuckDuckGo) for preselected sites.

oltarasenko:
That sounds like a super big and interesting project! I think it would be nice to hear the story!

PR #256, opened 2023-04-03
Improve UniqueRequest

serpent213:
* optional hashing (closes elixir-crawly/crawly#143)
* ignore trailing slash, assuming that pathes /a and /a/ yield identical content

### DIFF
--- a/test/middlewares/unique_request_test.exs
+++ b/test/middlewares/unique_request_test.exs
@@ -2,6 +2,7 @@ defmodule Middlewares.UniqueRequestTest do
   use ExUnit.Case, async: false
 
   @valid %Crawly.Request{url: "https://www.some_url.com"}
+  @valid_slash %Crawly.Request{url: "https://www.some_url.com/"}
 
   test "Filters out requests non-unique urls" do
     middlewares = [Crawly.Middlewares.UniqueRequest]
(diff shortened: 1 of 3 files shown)


test/middlewares/unique_request_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule Middlewares.UniqueRequestTest do
   use ExUnit.Case, async: false
 
   @valid %Crawly.Request{url: "https://www.some_url.com"}
```
serpent213, line 4: Actually, a valid hostname would not contain an underscore, I just noticed... 😉

oltarasenko:
@serpent213 It looks good from my point of view. However, I don't see any reason why we want to make it optional. E.g. I don't see why this middleware should store the URL in any case if it's better to store just a hash. Could you please make a switch so it uses hash by default and urls are not used at all anymore?

serpent213:
@oltarasenko My reasoning was the following:

* Correctness first – no need to even think about collisions when using the URL.
* Memory – let's say we use SHA256 as default, that would mean 32 bytes. Some URLs are shorter than that, some are longer, depends on the use case. So I thought it should be a conscious choice; also the possibility to choose e.g. MD5 if memory is more important than correctness.
* For most users it doesn't matter anyway, I imagine.

I could as well make it two different Middlewares, like UniqueRequest and HashedUniqueRequest.

oltarasenko (dismissed):
@serpent213 Ok, I think I agree with your points here. So I don't have any other comments!

oltarasenko:
Hey, @serpent213 could you please re-base and format your code so I can merge it. Thanks for the contribution!

serpent213:
> Hey, @serpent213 could you please re-base and format your code so I can merge it. Thanks for the contribution!

Looks good now? I hope, my git-fu didn't skip out on me.  😉

PR #261, opened 2023-05-20
Let Management Tool run in any place

aej:
This makes it possible to run the Management Tool nested inside other routers/plugs.

I've successfully tested this running it inside another router using `forward/2` (https://hexdocs.pm/plug/Plug.Router.html#forward/2).

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,18 @@ It allows to:
  - View/Download items extracted
  - View/Download logs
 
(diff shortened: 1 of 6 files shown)


oltarasenko:
@aej It looks good to me. Let me test it first and merge it afterwords.

oltarasenko (approved):

PR #267, opened 2023-07-16
Fix issue in the spider generator

Rukomoynikov:
Hello all.

I made this small PR to address the issue in the spider generator. From the documentation, I found that to generate a template for a spider, I should use:

```elixir
mix crawly.gen.spider --filepath ./lib/crawly_example/books_to_scrape.ex --spidername BooksToScrape
```

But if I change the order of parameters it will not work. The problem is that `OptParser` returns parsed arguments in a `key list` - `[filepath: filepath, spidername: spidername]`. So, for the function `defp generate_spider(filepath: filepath, spidername: spidername)`, it's necessary to pass it in this order (filepath first etc). If it's passed in another order, like spidername first, it fails. The proposed solution is to convert the key lists type into a Map. 

It could be solved without converting to a Map, adding a function:
```elixir 
defp generate_spider(spidername: spidername, filepath: filepath) do
  generate_spider(filepath: filepath, spidername: spidername)
end
# What if, in the future, more arguments will be added?
```

**Changes in the PR**
1. Fixed issue with the generator
2. Handling exceptions when a file with the spider can't be saved
3. Tests for spider generator

### DIFF
--- a/lib/mix/tasks/crawly.gen.spider.ex
+++ b/lib/mix/tasks/crawly.gen.spider.ex
@@ -33,11 +33,11 @@ defmodule Mix.Tasks.Crawly.Gen.Spider do
         help()
 
       true ->
-        generate_spider(opts)
+        Map.new(opts) |> generate_spider()
     end
   end
 
-  defp generate_spider(filepath: filepath, spidername: spidername) do
+  defp generate_spider(%{filepath: filepath, spidername: spidername}) do
     case File.exists?(filepath) do
       true ->
         Mix.shell().error("The spider already exists. Choose another filename")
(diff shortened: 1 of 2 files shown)


Rukomoynikov:
Hi @oltarasenko 

Could you please, when you have time, look at the PR? I can adjust it if you feel it doesn't fulfil its goal or remove it if the issue is not crucial.

oltarasenko:
Hey @Rukomoynikov,

Thanks for contributing, and sorry for a bit late reply, I have just returned from vacation. Will look at the PR soon.

oltarasenko (approved):

PR #279, opened 2023-11-19
Use Crawly.Engine.list_know_spiders instead of Crowly.running_spiders

cmnstmntmn:
According to [this comment here](https://github.com/elixir-crawly/crawly/blob/master/lib/crawly.ex#L129)

```ex
Crawly.Engine.list_know_spiders()
```

should be used instead of

```ex
Crowly.running_spiders()
```

### DIFF
--- a/lib/crawly/api.ex
+++ b/lib/crawly/api.ex
@@ -55,11 +55,12 @@ defmodule Crawly.API.Router do
 
   # Simple UI for crawly management
   get "/" do
+    known_spiders = Crawly.Engine.list_known_spiders()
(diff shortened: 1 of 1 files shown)


dogweather:
Thanks, @cmnstmntmn, This is interesting. I'm giving it a try.

kasvith:
Yes, it seems `Crawly.Engine.list_know_spiders()` hangs

oltarasenko:
Sorry, I was on another project for quite a bit of time. I have rebased your PR and will merge it as a part of https://github.com/elixir-crawly/crawly/pull/291

PR #282, opened 2023-12-13
Increase the error level of some messages

dogweather:
See https://github.com/elixir-crawly/crawly/discussions/270. In a nutshell, some errors are being logged as `debug`. This causes them to be hidden if the user is running Crawly as a normal user, at e.g. `info` or `warn` level.

* This PR increases the log level of these. 
* A few other logging statements are changed to hopefully better match the severity.
* E.g., some notifications were raised to `error` because they indicate a configuration error, and the app won't function as expected.

### DIFF
--- a/lib/crawly.ex
+++ b/lib/crawly.ex
@@ -141,7 +141,7 @@ defmodule Crawly do
       Crawly.Models.YMLSpider.load()
     rescue
       error ->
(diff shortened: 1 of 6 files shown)


dogweather:
All the tests passed. CI is failing due to a configuration problem.

oltarasenko (approved):

PR #287, opened 2024-02-26
[Fix] Resolve Protocol.UndefinedError by using fallback

OldhamMade:
Fixes an issue where `:requests` is `nil` but the following lines expect an empty list.

This was a quick fix for my own project, so I haven't put together any tests yet. However, all current tests pass locally.


lib/crawly/worker.ex
```diff
@@ -158,8 +158,8 @@ defmodule Crawly.Worker do
              parsed_item: Crawly.ParsedItem.t(),
              result: {:ok, :done}
   defp process_parsed_item({parsed_item, response, spider_name}) do
-    requests = Map.get(parsed_item, :requests, [])
-    items = Map.get(parsed_item, :items, [])
+    requests = Map.get(parsed_item, :requests) || []
```
oltarasenko, line 161: Sorry, I don't quite understand the difference between the:
`requests = Map.get(parsed_item, :requests) || []` and 
`requests = Map.get(parsed_item, :requests, [])`

OldhamMade:
```elixir
iex(1)> Map.get(%{requests: nil}, :requests, [])
nil
iex(2)> Map.get(%{requests: nil}, :requests) || []
[]
```

OldhamMade: Line 164 expects a list and fails on `nil`:

```elixir
iex(1)> Enum.each(nil, &(&1))
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom
...
iex(2)> Enum.each([], &(&1))
:ok
```

oltarasenko: Ah, ok. Clear now

mix.exs
```diff
@@ -2,7 +2,7 @@ defmodule Crawly.Mixfile do
   use Mix.Project
 
   @source_url "https://github.com/oltarasenko/crawly"
-  @version "0.16.0"
+  @version "0.16.1"
```
oltarasenko, line 5: Please remove this change

oltarasenko (changes requested):
Please rebase your commit and remove changes to mix.exs, so I can merge the code

OldhamMade:
Done!

oltarasenko (approved):

PR #295, opened 2024-04-29
Add normalise_url parameter to Crawly.Middlewares.UniqueRequest

adonig:
This parameter allows to customize the normalization behavior using a unary normalization function.

### DIFF
--- a/lib/crawly/middlewares/unique_request.ex
+++ b/lib/crawly/middlewares/unique_request.ex
@@ -17,6 +17,12 @@ defmodule Crawly.Middlewares.UniqueRequest do
   ]
   ```
 
+  ```
+  middlewares: [
+    {Crawly.Middlewares.UniqueRequest, hash: :sha, normalise_url: fn url -> String.trim_trailing("/") end}
+  ]
+  ```
+
   See the [Erlang documentation for crypto](https://www.erlang.org/doc/man/crypto.html#type-sha1)
   for available algorithms.
   """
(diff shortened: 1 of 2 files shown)


oltarasenko:
Sorry, I don't understand the idea behind this change. Could you please explain?

adonig:
It allows you to use a different normalization function without having to replace the whole UniqueRequest middleware. For instance, while some search engines consider URLs with and without a trailing slash as identical, others do not. Some go even further and treat /index.html the same way. Some transform the host part of the URL to lowercase and some even sort the query string parameters alphanumerically or resolve relative paths. For example I use this normalization function:

```elixir
  def normalize_url(url) do
    parsed = URI.parse(url)

    if parsed.scheme in ["http", "https"] and parsed.host do
      %URI{
        scheme: parsed.scheme,
        host: String.downcase(parsed.host),
        path: parsed.path,
        query: parsed.query
      }
      |> URI.to_string()
    else
      nil
    end
  end
```

adonig:
[Section 6 of RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6) goes a bit deeper into the topic of URL normalization.

adonig:
I found out that Erlang comes with a RFC 3986-compliant URL normalization function: [:uri_string.normalize/1](https://www.erlang.org/doc/man/uri_string#normalize-1)

I believe it's still a good idea to allow people to provide their own implementation, because some might want to extend the behavior of the RFC, like for example [Cloudflare](https://developers.cloudflare.com/rules/normalization/how-it-works/) or [Kaspersky](https://support.kaspersky.com/CyberTrace/4.4/en-US/231546.htm) do.

dogweather:
@adonig Do you use this different behavior when you use Crawly?

oltarasenko (approved):
This change looks good to me. Sorry I somehow missed the PR :(

